PR #1, opened 2021-02-15
Clone from jenkins-jobs repository for job updates

ehelms:


ekohl (changes requested):
Related we also have [duffy.groovy](https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/lib/duffy.groovy) and [foreman_infra.groovy](https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/lib/foreman_infra.groovy). For this job you should also update this line (and maybe renamed the whole job/file):
https://github.com/theforeman/jenkins-jobs/blob/fa60d415a7b30d981a79c800caf0b64ac9e89c13/theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml#L7

ehelms:
Got all the other locations covered now

theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
```diff
@@ -4,7 +4,7 @@
     sandbox: true
     properties:
       - github:
-          url: https://github.com/theforeman/foreman-infra
+          url: https://github.com/theforeman/jenkins-jobs
```
ekohl, line 7: It feels odd to keep this job named `foreman-infra` while the job has been changed to a different repo.

theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -2,7 +2,7 @@ void git_clone_foreman_infra(args = [:]) {
     target_dir = args.target_dir ?: ''
```
ekohl, line 2: It's odd to name it `git clone foreman-infra` and then you end up with `jenkins-jobs`

theforeman.org/pipelines/lib/duffy.groovy
```diff
@@ -1,8 +1,8 @@
 def provision() {
     fix_ansible_config()
-    git_clone_foreman_infra(target_dir: 'foreman-infra')
+    git_clone_foreman_infra(target_dir: 'jenkins-jobs')
 
-  	dir('foreman-infra/ci/centos.org/ansible') {
+  	dir('centos.org/ansible') {
```
ekohl, line 5: I think this needs to continue using the target dir in the path

theforeman.org/pipelines/lib/duffy.groovy
```diff
@@ -15,19 +15,19 @@ def fix_ansible_config() {
 }
 
 def deprovision() {
-    if (fileExists('foreman-infra')) {
-        dir('foreman-infra/ci/centos.org/ansible') {
+    if (fileExists('jenkins-jobs')) {
+        dir('centos.org/ansible') {
```
ekohl, line 19: Same here

centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -62,7 +62,7 @@ pipeline {
                 extra_vars = buildExtraVars(extraVars: playBook['extraVars'])
                 duffy_ssh("cd forklift && ansible-playbook playbooks/collect_debug.yml --limit '${playBook['boxes'].join(',')}' ${extra_vars}", 'duffy_box', './')
                 runPlaybook(
-                    playbook: 'foreman-infra/ci/centos.org/ansible/fetch_debug_files.yml',
+                    playbook: 'centos.org/ansible/fetch_debug_files.yml',
```
ekohl, line 65: I think this also needs to contain the directory name

theforeman.org/yaml/jobs/foreman-infra-jenkins-jobs-update.yml
ehelms: The intent was not to keep it that way long term, I have been trying to do this in stages this whole time to reduce the amount of places we might break something.

theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -2,7 +2,15 @@ void git_clone_foreman_infra(args = [:]) {
...
 
     dir(target_dir) {
-        git url: 'https://github.com/theforeman/foreman-infra', poll: false
+        git url: 'https://github.com/theforeman/foreman_infra', poll: false
+    }
+}
+
+void git_clone_jenkins_jobs(args = [:]) {
```
ehelms, line 9: Added a new method for this repository given that some jobs need the puppet part of foreman-infra.

ekohl (approved):

PR #5, opened 2021-02-19
Add Ruby 2.7 testing to installer jobs

ehelms:


theforeman.org/pipelines/release/source/foreman-installer.groovy
```diff
@@ -22,6 +22,11 @@ pipeline {
                 run_test(ruby: '2.5', puppet: '6.3')
             }
         }
+        stage("test ruby 2.7 & puppet 6") {
+            steps {
+                run_test(ruby: '2.7', puppet: '6.3')
```
ekohl, line 27: Perhaps it should be Puppet 7 since that ships with Ruby 2.7?

ekohl (changes requested):

theforeman.org/pipelines/release/source/foreman-installer.groovy
ehelms: I thought the ruby version was supposed to reflect the installer Ruby runtime, and puppet version should be the one we most often use to catch the most issues. I can add two new entries:  2.7+P7, 2.7+P6

ekohl: I don't really think that's worth it and doubt it can be done. Puppet is installed as a gem and thus runs under the Ruby env.

I have it on my agenda to support Puppet 7 in all modules and I hope to switch nightly to it soon after.

ehelms:
How does this matrix look?

ekohl (commented):
Can you explain why you no longer use `PUPPET_VERSION`? AFAIK this is still needed to actually be effective due to https://github.com/theforeman/foreman-installer/blob/8310aa688aefd6a339fa7589834b89d79845c151/Gemfile#L5

ehelms:
> Can you explain why you no longer use `PUPPET_VERSION`? AFAIK this is still needed to actually be effective due to https://github.com/theforeman/foreman-installer/blob/8310aa688aefd6a339fa7589834b89d79845c151/Gemfile#L5

Simple forgetting that there is an implied environment variable created that the code itself expects.

ekohl:
In my mind I always start from the bottom. That means going over all modules to ensure they support Puppet 7. Then we can update the check for Puppet 7 here:
https://github.com/theforeman/foreman-installer/blob/8310aa688aefd6a339fa7589834b89d79845c151/spec/puppet_version_spec.rb#L5

I've submitted a bunch of PRs to test on Puppet 7 now:
* https://github.com/theforeman/puppet-candlepin/pull/178
* https://github.com/theforeman/puppet-certs/pull/319
* https://github.com/theforeman/puppet-dhcp/pull/186
* https://github.com/theforeman/puppet-dns/pull/184
* https://github.com/theforeman/puppet-foreman/pull/921
* https://github.com/theforeman/puppet-foreman_proxy/pull/652
* https://github.com/theforeman/puppet-foreman_proxy_content/pull/333
* https://github.com/theforeman/puppet-git/pull/74
* https://github.com/theforeman/puppet-katello/pull/392
* https://github.com/theforeman/puppet-katello_devel/pull/254
* https://github.com/theforeman/puppet-motd/pull/4
* https://github.com/theforeman/puppet-pulp/pull/412
* https://github.com/theforeman/puppet-pulpcore/pull/174
* https://github.com/theforeman/puppet-puppet/pull/779
* https://github.com/theforeman/puppet-puppetserver_foreman/pull/5
* https://github.com/theforeman/puppet-qpid/pull/147
* https://github.com/theforeman/puppet-tftp/pull/111

We should also go over the dependencies and check where we need to bump dependencies.

ehelms:
Please keep in mind I am not intending to add Puppet 7 support here. I had added Ruby 2.7 support and was informed that I should also do so via pairing it with Puppet 7. I want to test and ensure that both Kafo and the Installer's Ruby code runs under 2.7 (as it would on an RPM install) not testing Puppet's runtime ruby.

theforeman.org/pipelines/test/kafo.groovy
```diff
@@ -11,11 +11,33 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '2.4', '2.5', '2.6'
```
ekohl, line 14: If we drop 2.4, then IMHO it should also be reflected in the gemspec. Also a PR to drop Puppet 5 support would be the correct thing to do.

ehelms:
I'll keep Puppet 5 around. This uses `notValues` in the excludes which to my understanding should mean, even though using `excludes` that what you see in the section represents the valid combination of values.

ekohl (approved):

PR #12, opened 2021-03-03
unify katello pipeline definitions

evgeni:
let's generate el8 upgrade pipes for everyone


ekohl (approved):
Looks like it was already generating el8 pipelines for all and this effectively the same as we already did.

evgeni:
no, *some* el8 weren't generated before that, but it doesn't hurt to generate them anyways

ekohl:
Ah yes. Exactly the one that made the least sense to skip. If it was 3.x on EL8 upgrades then I would have understood it ...

ehelms:
Why do we want to generate jobs that don't make sense? Thats cruft and confusion. I'd prefer we unify things when things are actually unified, not artifically.

evgeni:
Because it leads to things like this, where we thought the job existed, but it did not? :)

This PR just adds the one job that was actually missing.

For the EL8 upgrade jobs for pre-4.0: :shrug:, I never look at ci.centos.org directly anyways, so it never felt like confusion. and ci.tfm.o knows which jobs it shall trigger.

ehelms:
Well, I'd argue that this commit https://github.com/theforeman/jenkins-jobs/commit/fb3152135328666c7328f554950ebffaf2741da8# was incorrect and hence why we thought it might should just work.

For me, this incorrectly specifies the state of jobs and what we support. I tend to not like that sort of incorrectness because it leads to confusion when looking at it now and 3 months down the line. I don't mind being overridden here, and I know I at least will be confused sometime in the near future by this.

evgeni:
I want to see that bridge. Are you aiming at "when we add stream8"?

ehelms:
Honestly, this matrix is always going to ebb and flow. As we add stream 8 and drop centos 8, as we add stream 9, as we drop EL7. I just prefer correctness of that matrix over trying to simplify the definition within this file.

evgeni:
I think the "difference" is that I don't see *this* part of the job definition as the matrix, but this:

https://github.com/theforeman/jenkins-jobs/blob/216157a7c916d207306256c4fc2fb541c852834c/theforeman.org/pipelines/vars/katello/nightly.groovy#L6-L15

The stuff in `centos.org` is just "whatever needs to get the job done"

evgeni:
so instead you want something like https://github.com/theforeman/jenkins-jobs/pull/13?

ehelms:
That is my preference, I won't hold up this approach either, I just can't ACK it :)

PR #13, opened 2021-03-04
exactly define what katello jobs are required

evgeni:


ehelms (approved):
Two roads diverged in a wood....

PR #16, opened 2021-03-08
Add testing Ruby 2.7 for Katello

ehelms:
Changes to a matrix style job for Katello PR testing and drops
the previous parallelism.

After switching packaging we can drop 2.5 for nightly. I have observed that the Ruby 2.7 tests run 15-20 minutes faster than 2.5. I think the previous parallelism was incorrect and not providing any benefit. Further, you cannot nest parallel instead matrix. Talking with Katello team, they are OK going this route for a while and seeing how things perform.

### DIFF
--- a/theforeman.org/pipelines/test/testKatello.groovy
+++ b/theforeman.org/pipelines/test/testKatello.groovy
@@ -1,175 +1,158 @@
 def katello_versions = [
     'master': [
         'foreman': 'develop',
-        'ruby': '2.5',
+        'ruby': ['2.5', '2.7']
     ],
     'KATELLO-4.0': [
         'foreman': '2.4-stable',
-        'ruby': '2.5'
+        'ruby': ['2.5']
     ],
     'KATELLO-3.18': [
         'foreman': '2.3-stable',
-        'ruby': '2.5'
+        'ruby': ['2.5']
     ],
     'KATELLO-3.17': [
         'foreman': '2.2-stable',
-        'ruby': '2.5'
+        'ruby': ['2.5']
     ],
 ]
 
-def ruby = katello_versions[ghprbTargetBranch]['ruby']
+def ruby_versions = katello_versions[ghprbTargetBranch]['ruby']
 def foreman_branch = katello_versions[ghprbTargetBranch]['foreman']
 
 pipeline {
+    agent none
+
     options {
         timestamps()
         timeout(time: 3, unit: 'HOURS')
         ansiColor('xterm')
     }
 
-    agent { label 'fast' }
-
     stages {
-        stage('Setup Git Repos') {
-            steps {
-                deleteDir()
-                ghprb_git_checkout()
-
-                dir('foreman') {
-                   git url: "https://github.com/theforeman/foreman", branch: foreman_branch, poll: false
-                }
-            }
-        }
-        stage("Setup RVM") {
-            steps {
-
-                configureRVM(ruby)
-
-            }
-        }
-        stage('Configure Environment') {
-            steps {
-
-                dir('foreman') {
-                    addGem()
-                    databaseFile(gemset())
-                }
-
-            }
-        }
-        stage('Configure Database') {
-            steps {
-
-                dir('foreman') {
-                    configureDatabase(ruby)
-                }
-
-            }
-        }
-        stage('Install Foreman npm packages') {
-            steps {
-                dir('foreman') {
-                    withRVM(["bundle exec npm install"], ruby)
-                }
-            }
-        }
-        stage('Run Tests') {
-            parallel {
-                stage('tests') {
-                    steps {
-                        dir('foreman') {
-                            withRVM(['bundle exec rake jenkins:katello TESTOPTS="-v" --trace'], ruby)
-                        }
+        stage('Test') {
+            matrix {
+                agent { label 'fast' }
+                axes {
+                    axis {
+                        name 'ruby'
+                        values '2.5', '2.7'
                     }
                 }
-                stage('rubocop') {
-                    steps {
-                        dir('foreman') {
-                            withRVM(['bundle exec rake katello:rubocop TESTOPTS="-v" --trace'], ruby)
-                        }
+                when {
+                    expression {
+                        ruby_versions.contains(env.ruby)
                     }
                 }
-                stage('react-ui') {
-                    when {
-                        expression { fileExists('package.json') }
-                    }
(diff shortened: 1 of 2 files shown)


theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,187 @@
...
+        'ruby': ['2.5']
     ],
 ]
 
-def ruby = katello_versions[ghprbTargetBranch]['ruby']
-def foreman_branch = katello_versions[ghprbTargetBranch]['foreman']
+//def ruby_versions = katello_versions[ghprbTargetBranch]['ruby']
+//def foreman_branch = katello_versions[ghprbTargetBranch]['foreman']
+def ruby_versions = katello_versions['master']['ruby']
+def foreman_branch = katello_versions['master']['foreman']
```
evgeni, line 23: this doesn't look right? ;)

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,187 @@
...
-
-                                    script {
-                                        bastion_install = readFile('bastion-version')
-                                        bastion_version = bastion_install.split('bastion-')[1]
-                                        echo bastion_install
-                                        echo bastion_version
+                    stage('angular-ui') {
+                        steps {
+                            script {
+                                if (!fileExists('engines/bastion')) {
```
evgeni, line 112: bastion was merged in 3.12, so it's safe to say this will always be "false" now

ekohl (commented):
https://github.com/Katello/katello/blob/62a79f9ca909c073d285f5775a4f3c97c44cbb00/katello.gemspec#L16 claims it's not compatible with 2.7. I pointed this out to Katello devs. Probably not a blocker, but something that isn't very nice.

theforeman.org/pipelines/test/testKatello.groovy
ehelms: Truth, I am still testing this. Shoulda gone for draft mode since Katello has a failing test.

ehelms:
This is part of enabling that for Katello -- https://github.com/Katello/katello/pull/9185

theforeman.org/pipelines/test/testKatello.groovy
ehelms: Tackled.

ehelms:
OK I went ahead and updated this back to PR testing mode as I think we have a resolution on the Katello code side.

ehelms:
I think as a follow up change, I will look at splitting the JS out into its own test job all together. This will allow each type of testing, JS or Ruby, to run independently and allow future matrix of NodeJS versions when we want to move to NodeJS 14.

ehelms:
@evgeni Mind taking a look at this again?

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,163 @@
...
+                        values '2.5', '2.7'
+                    }
                 }
-            }
-        }
-        stage("Setup RVM") {
-            steps {
+                when {
+                    expression {
+                        ruby_versions.contains(env.ruby)
```
ekohl, line 44: Is this set up because you can't use variables in `values`?

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,163 @@
...
+                            }
+                        }
                     }
-                    steps {
-                        sh "npm install"
-                        sh 'npm test'
+                    stage('react-ui') {
+                        when {
+                            expression { fileExists('package.json') }
+                        }
```
ekohl, line 100: If you dropped it for angular-ui, I think you can also drop it here.

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,163 @@
...
-                stage('react-ui') {
-                    when {
-                        expression { fileExists('package.json') }
+                    stage('rubocop') {
+                        steps {
+                            dir('foreman') {
+                                withRVM(['bundle exec rake katello:rubocop TESTOPTS="-v" --trace'], ruby)
+                            }
+                        }
                     }
```
ekohl, line 96: Maybe not something you want to change in this PR, but I'd recommend running Rubocop only once (on the latest version). Also consider publishing the results like we do in smart-proxy:
https://github.com/theforeman/jenkins-jobs/blob/ed25297a67807f6d5684c57ab42cd6fde75a8627/theforeman.org/pipelines/test/smart-proxy.groovy#L35-L43

Note that it takes the approach of running rubocop in the tool itself rather than via the common rake job via Foreman which may or may not be what you want.

I'd also run rubocop before running `npm install` so you fail fast. In fact, I would suggest to run rubocop at the very beginning so you execute as little as possible if rubocop fails.

theforeman.org/pipelines/test/testKatello.groovy
ehelms: Correct.

theforeman.org/pipelines/test/testKatello.groovy
ehelms: That sounds like a good follow up given I need to make sure it all works and I am contemplating some further improvements to break out the design.

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -1,175 +1,160 @@
...
+                }
+                post {
+                    always {
+                        dir('foreman') {
+                            archiveArtifacts artifacts: "log/test.log"
+                            junit keepLongStdio: true, testResults: 'jenkins/reports/unit/*.xml'
+                        }
+                        dir('foreman') {
+                            cleanup(ruby)
+                        }
```
ekohl, line 152:
```suggestion
                        dir('foreman') {
                            archiveArtifacts artifacts: "log/test.log"
                            junit keepLongStdio: true, testResults: 'jenkins/reports/unit/*.xml'
                            cleanup(ruby)
                        }
```

ekohl (approved):
I wonder if you test bastion and bastion_katello in parallel and how much benefit that would give.

ehelms:
> I wonder if you test bastion and bastion_katello in parallel and how much benefit that would give.

Due to the matrix wrapper, nested parallel cannot be used. I am thinking of splitting JS out to it's own job to make it faster and independently runnable.

PR #18, opened 2021-03-09
Install collections when setting up Forklift

ehelms:


evgeni (approved):
ACK pending it's actually needed ;)

evgeni:
but also see my last idea in https://github.com/theforeman/forklift/pull/1313#issuecomment-814956283

evgeni (approved):

PR #19, opened 2021-03-14
Add rh_cloud plugin to jenkins PR tests

ShimShtein:
This is a first attempt, would like to hear how to do it in a more proper way


evgeni:
we use GHA now, closing

PR #23, opened 2021-03-19
push katello RPMs to fedorapeople AND yum.tfm

evgeni:


evgeni:
this should push ka

ehelms (approved):
Let's give that a try, it'll allow testing and then we can consider updating the release repos and docs.

PR #25, opened 2021-03-22
correct katello paths

evgeni:
also keep old files only for releases

### DIFF
--- a/theforeman.org/pipelines/lib/release.groovy
+++ b/theforeman.org/pipelines/lib/release.groovy
@@ -31,8 +31,8 @@ void push_pulpcore_rpms(version, distro) {
 
 void push_katello_rpms(version, distro) {
     keep_old = version != 'nightly'
-    push_rpms("katello-${version}/katello", "katello/katello", version, distro, keep_old)
-    push_rpms("katello-${version}/candlepin", "katello/candlepin", version, distro, keep_old)
+    push_rpms_direct("katello-${version}/katello/${distro}", "katello/${version}/katello/${distro}", !keep_old, keep_old)
+    push_rpms_direct("katello-${version}/candlepin/${distro}", "katello/${version}/candlepin/${distro}", !keep_old, keep_old)
 }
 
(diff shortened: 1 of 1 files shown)


ekohl (approved):

theforeman.org/pipelines/lib/release.groovy
```diff
@@ -30,7 +30,9 @@ void push_pulpcore_rpms(version, distro) {
 }
 
 void push_katello_rpms(version, distro) {
-    push_rpms("katello-${version}", "katello", version, distro, true)
+    keep_old = version != 'nightly'
+    push_rpms("katello-${version}/katello", "katello/katello", version, distro, keep_old)
+    push_rpms("katello-${version}/candlepin", "katello/candlepin", version, distro, keep_old)
```
evgeni, line 35: this will result in things being put into `https://yum.theforeman.org/katello/katello/<version>` and `https://yum.theforeman.org/katello/candlepin/<version>`

not the "nicest" path structure, but predictable.

I could also try using `push_rpms_direct` which allows more fexibility…

evgeni: on the old host, it's `https://fedorapeople.org/groups/katello/releases/yum/<version>/<component>`

ekohl (commented):
Btw, I've started to copy the Linux kernel and use "Fixes: f0d939ada7b9d4e380ba058d17e65376a1408a99" in the commit message when I fix a specific commit.

theforeman.org/pipelines/lib/release.groovy
ehelms: I do kinda like `<version>/component` given Candlepin is not stand-alone and if you are browsing it gives more of view that you need both paths. If thats harder to do though, this will be fine.

evgeni:
@ekohl @ehelms updated to use push_rpms_direct with nicer paths

ehelms (approved):

PR #27, opened 2021-03-23
cleanup ci.centos.org better

evgeni:


ekohl (approved):
If there's a cleanup step, we should use it a lot more. We call `deleteDir` and RVM cleanup in many places.

PR #30, opened 2021-03-24
Switch kafo and installer to pure excludes matrix for Puppet and Ruby…

ehelms:
… versions

The use of notValues for both slots of the matrix created no viable
job configurations. One part would need to be notValues and the other
values in order to work. Given this can be confusing to read, this
opts for a more classic exclusion model pairing up ruby and puppet
versions that should not be tested together explicitly.

Fixes c94c33fb1898600fd499240e932f709d6ccfec74

### DIFF
--- a/theforeman.org/pipelines/test/foreman-installer.groovy
+++ b/theforeman.org/pipelines/test/foreman-installer.groovy
@@ -27,7 +27,7 @@ pipeline {
                         }
                         axis {
                             name 'PUPPET_VERSION'
-                            notValues '5.0'
+                            values '5.0'
                         }
                     }
                     exclude {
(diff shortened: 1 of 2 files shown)


theforeman.org/pipelines/test/foreman-installer.groovy
```diff
@@ -23,31 +23,31 @@ pipeline {
                     exclude {
                         axis {
                             name 'ruby'
-                            notValues '2.4'
+                            values '2.5', '2.7'
```
ekohl, line 26: Looking at this, shouldn't it exclude `notValues '2.4'` and `values '5.0'`.

ehelms: Alright. Changed it to that. I will screw this up again in the future :)

ehelms:
Example run of this change -- https://ci.theforeman.org/blue/organizations/jenkins/kafo-pr-test/detail/kafo-pr-test/276/pipeline/

ekohl (approved):

PR #34, opened 2021-03-26
use cleanup step in source pipelines

evgeni:
this is just cosmetics to align them with the other pipes

### DIFF
--- a/theforeman.org/pipelines/release/source/foreman.groovy
+++ b/theforeman.org/pipelines/release/source/foreman.groovy
@@ -37,8 +37,10 @@ pipeline {
                     }
                     post {
                         always {
-                            cleanup(env.RUBY_VER, env.GEMSET)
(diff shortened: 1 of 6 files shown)


theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -37,8 +37,10 @@ pipeline {
                     }
                     post {
                         always {
-                            cleanup(env.RUBY_VER, env.GEMSET)
                             junit(testResults: 'jenkins/reports/unit/*.xml')
+                        }
+                        cleanup {
+                            cleanup(env.RUBY_VER, env.GEMSET)
```
ekohl, line 43: This looks confusing. Perhaps our `cleanup` should really be `rvmCleanup` or similar. However, that's for another day.

ekohl (approved):

PR #38, opened 2021-04-12
start building foreman core on focal

evgeni:


ekohl:
Don't we first need to build all dependencies? I recall that last time the issue was there. However, by now we've fixed Kafo. Or did you already rebuild those?

evgeni:
Yes, all deps are built on focal :)

ekohl (approved):

PR #39, opened 2021-04-12
add focal to debian pipelines

evgeni:
draft as I don't want to enable them just yet


ekohl (approved):
IMHO the last commit can already be merged.

PR #43, opened 2021-04-15
add centos8 stream to foreman nightly pipelines

evgeni:


ekohl (approved):
Technically I'd expect 2.4 to also pass and 2.3 to fail due to missing SELinux and installer fixes but let's start with nightly.

PR #44, opened 2021-04-15
add centos8 stream to katello nightly

evgeni:


evgeni:
This now happened. Yay

PR #46, opened 2021-04-19
allow building non-gem dependencies

evgeni:


theforeman.org/scripts/debian/execute_pbuilder_dependency.sh
```diff
@@ -12,10 +12,19 @@ if [ -e ../${project}/build_vars.sh ]; then
   . ../${project}/build_vars.sh
 fi
 
-# Figure out package version and gem2deb it
+# Figure out package version and get it
 VERSION=$(head -n1 ../${project}/changelog|awk '{print $2}'|sed 's/(//;s/)//'|cut -f1 -d-|cut -d: -f2)
-gem fetch ${project} -v "=${VERSION}"
-../../gem2deb ${project}-${VERSION}.gem --debian-subdir ../${project} --only-source-dir
+if [[ ${project} = python-* ]]; then
+  PIP_NAME=$(echo ${project}| sed '/^python-//')
```
ekohl, line 18:
```suggestion
  PIP_NAME=${project#python-}
```

evgeni:
reworked to load BUILD_TYPE from a file, instead of guessing.

theforeman.org/scripts/debian/execute_pbuilder_dependency.sh
```diff
@@ -7,15 +7,28 @@ cd dependencies/${os}
...
 if [ -e ../${project}/build_vars.sh ]; then
   . ../${project}/build_vars.sh
 fi
 
-# Figure out package version and gem2deb it
+# Figure out package version and get it
 VERSION=$(head -n1 ../${project}/changelog|awk '{print $2}'|sed 's/(//;s/)//'|cut -f1 -d-|cut -d: -f2)
-gem fetch ${project} -v "=${VERSION}"
-../../gem2deb ${project}-${VERSION}.gem --debian-subdir ../${project} --only-source-dir
+if [ "${BUILD_TYPE}" == "python" ]; then
```
ekohl, line 18: I think `==` is a bashism. If you do that, isn't this even better?
```suggestion
if [[ $BUILD_TYPE == "python" ]]; then
```

evgeni: Gah, yes it is. And given the script is bash anyways, [[ is faster, being a built-in

evgeni:
the code duplication between the two build scripts is already high, and I didn't want to introduce a refactoring by using a shared lib just yet.

ekohl (approved):
I do wonder if we should go full on with the duplication and have the same code twice even if unused. Then at some point we can maybe merge the two scripts into one by looking at an argument to determine whether it's a dependency or plugin.

PR #48, opened 2021-04-26
install pbr before installing jjb

evgeni:
for some reason, pip fails to install them if that happens in the same
transaction


ekohl (commented):
Could it be updating setuptools this way?

evgeni:
seems it was some pypi hickup -- it was reproducible earlier, and isn't anymore

greeeeeeeeeeeeeat

PR #51, opened 2021-05-03
Add foreman 2.5 jobs

zjhuntin:


theforeman.org/pipelines/vars/foreman/2.5.groovy
```diff
@@ -0,0 +1,40 @@
+def foreman_version = '2.5'
```
ehelms, line 1: I think we want to include stream and maybe Ubuntu 2004 here. Let's check with @evgeni to be sure though

zjhuntin: After looking at the forklift versions, I think that you're right. I went ahead and updated this to include those two, but if they are not needed happy to remove them too.

theforeman.org/pipelines/vars/foreman/2.5.groovy
```diff
@@ -0,0 +1,44 @@
...
+    'el7',
+    'el6',
+    'sles12',
+    'sles11'
+]
+def foreman_el_releases = [
+    'el7',
+    'el8'
+]
+def foreman_debian_releases = ['buster', 'bionic']
```
evgeni, line 13:
```suggestion
def foreman_debian_releases = ['buster', 'bionic', 'focal']
```

theforeman.org/pipelines/vars/foreman/2.5.groovy
```diff
@@ -0,0 +1,44 @@
...
+def pipelines_deb = [
+    'install': [
+        'debian10',
+        'ubuntu1804',
+        'ubuntu2004',
+    ],
+    'upgrade': [
+        'debian10',
+        'ubuntu1804'
+        'ubuntu2004',
```
evgeni, line 24:
```suggestion
```

We don't have anything to upgrade from.

theforeman.org/pipelines/vars/foreman/2.5.groovy
```diff
@@ -0,0 +1,44 @@
...
+def pipelines_el = [
+    'install': [
+        'centos7',
+        'centos8',
+        'centos8-stream',
+    ],
+    'upgrade': [
+        'centos7',
+        'centos8',
+        'centos8-stream',
```
evgeni, line 37:
```suggestion
```

We don't have anything to upgrade from.

theforeman.org/yaml/jobs/packaging_build_deb_dependency.yaml
```diff
@@ -29,7 +29,7 @@
           values:
           - debian
     execution-strategy:
-      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.4" || version == "2.3") && os == "focal")'
+      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.5" || version == "2.4" || version == "2.3") && os == "focal")'
```
evgeni, line 32: no, focal is supported on 2.5 :)

```suggestion
      combination-filter: '(onlyos == "all" || onlyos == os) && !((version == "2.4" || version == "2.3") && os == "focal")'
```

evgeni (changes requested):

theforeman.org/pipelines/vars/foreman/2.5.groovy
ekohl: As a general note, starting from nightly rather than 2.4 is always a good idea. The only difference then would be `s/nightly/2.5/` and drop `fc29` from the client distros.

zjhuntin: Ack, I'm gonna go ahead and do that

evgeni (changes requested):
needs adding to https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/yaml/includes/foreman_versions.yaml.inc

theforeman.org/yaml/jobs/test_2_5_stable.yaml
```diff
@@ -0,0 +1,40 @@
...
+    triggers:
+      - scm_fifteen_minutes
+      - github
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+            - 2.5
+            - 2.6
```
evgeni, line 19: I think you can drop ruby 2.6 here, as we have no tergets using that.

@tbrisker was OK with that I think

theforeman.org/yaml/jobs/test_proxy_2_5_stable.yaml
```diff
@@ -0,0 +1,30 @@
...
+    triggers:
+      - scm_fifteen_minutes
+      - github
+    axes:
+      - axis:
+          type: user-defined
+          name: ruby
+          values:
+          - 2.5
+          - 2.6
```
evgeni, line 19: same

tbrisker (approved):
lgtm, we should probably also drop ruby 2.6 from the nightly ones if it's still there (but doesn't have to be in this pr)

evgeni:
> lgtm, we should probably also drop ruby 2.6 from the nightly ones if it's still there (but doesn't have to be in this pr)

Yeah, I would drop it in nightly separately.

evgeni (approved):

PR #53, opened 2021-05-04
drop ruby 2.6 from test matrix

evgeni:


evgeni:
I did not drop it from the 2.3 and 2.4 stable tests, just not to "change history"

tbrisker (approved):
Thanks @evgeni !

PR #54, opened 2021-05-04
Add Katello 4.1 + Foreman 2.5 mapping

jeremylenz:


theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -3,14 +3,14 @@ def katello_versions = [
...
+        'ruby': ['2.7']
+    ],
     'KATELLO-4.0': [
         'foreman': '2.4-stable',
         'ruby': ['2.5']
     ],
-    'KATELLO-3.18': [
-        'foreman': '2.3-stable',
-        'ruby': ['2.5']
-    ],
```
ekohl, line 13: I'd keep this in, at least until we really drop 3.18. It doesn't hurt to have a slightly bigger table.

ekohl (changes requested):

theforeman.org/pipelines/test/testKatello.groovy
evgeni: 3.18 shouldn't be dropped until 4.1 is actually GA

jeremylenz: Thanks, updated to restore 3.18.

evgeni (approved):

ekohl (approved):

PR #67, opened 2021-05-19
stop including psych as a gem

jlsherrill:


ehelms (approved):
Cleaning up our messes!

PR #70, opened 2021-05-20
remove further gem updates

jlsherrill:


evgeni (approved):

theforeman.org/scripts/test/test_hammer_cli_foreman_discovery.sh
```diff
@@ -6,7 +6,6 @@
 gemset=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
 rvm use ruby-${ruby}@${gemset} --create
 rvm gemset empty --force
-gem update --no-document
 gem install bundler -v '< 2.0' --no-document
```
ekohl, line 9: I wonder if by now we can't drop this as well.

jlsherrill: does rvm install bundler by default now? i imagine we don't need to specify '< 2.0' anymore even if it doesn't

PR #72, opened 2021-06-17
Rename skip-koji-whitelist-check to skip-koji-tag-check

ehelms:
See https://github.com/theforeman/obal/pull/293


ehelms:
We can either coordinate this merge change with the change in Obal or pin Obal ahead of time to a commit or version we check out.

ekohl:
We no longer have Koji, so that's another approach I guess.

PR #74, opened 2021-07-13
drop pulpcore 3.13 jobs, nobody is using that branch

evgeni:


ekohl (approved):
Thoughts about adding a README to https://yum.theforeman.org/pulpcore/3.13/ (and other "EOL" versions) that it's no longer maintained?

evgeni:
> Thoughts about adding a README to https://yum.theforeman.org/pulpcore/3.13/ (and other "EOL" versions) that it's no longer maintained?

We don't have that for foreman and katello, why would pulpcore be different?

That said, we could implement something like that for all projects?

ekohl:
The reason I asked was that with normal releases it's all linear. You know 2.4 & 2.5 are supported but everything older isn't. Here you have 3.11 & 3.14 supported, but 3.13 in between isn't. That surprising.

evgeni:
Yes, but also these repos aren't supposed to be used outside of the Katello context, and Katello knows which repo to enable.

ekohl:
Perhaps it's then sufficient to add a HEADER.html to https://yum.theforeman.org/pulpcore/ with that info.

ekohl:
https://github.com/theforeman/foreman-infra/pull/1624 should do so.

PR #77, opened 2021-07-22
Update pipeline to expect 3.0

tbrisker:


evgeni:
Interestingly, we did something to the nightly pipelines, that they stopped checking the expected version.

It still works fine on non-nightly, so :woman_shrugging:

evgeni:
Opened https://github.com/theforeman/jenkins-jobs/pull/78 and https://github.com/theforeman/tool_belt/pull/384 to drop this step

PR #78, opened 2021-07-23
unset expected version

evgeni:
we either pass the correct one (in the release jobs) or an empty string
(in the nightly jobs), so there is really no need to keep an artificial
nightly number in here, that is never checked


evgeni:
for the curious, I think the change in 986095c9087b1ff23455e9eea6fc70a152a69866 made it always pass an empty string, if the version is unset (as it is on nightly), so the default isn't used anymore.

ehelms (approved):

PR #80, opened 2021-07-23
copy null db schema for foreman

jlsherrill:


jlsherrill:
this isn't strictly needed, as the current tests run, but its likely they run by chance due to other tasks being run

theforeman.org/scripts/test/test_develop.sh
```diff
@@ -54,6 +54,8 @@ bundle exec rake $tasks TESTOPTS="-v" --trace
 
 # Test asset precompile
 if [ "${UI}" = "true" ]; then
+  rm db/schema.rb
```
ehelms, line 57:
```suggestion
  rm -f db/schema.rb
```

ehelms (approved):

PR #81, opened 2021-08-02
Foreman 3.0 jobs

pcreech:


theforeman.org/yaml/includes/foreman_versions.yaml.inc
```diff
@@ -1,3 +1,3 @@
-- '2.3'
 - '2.4'
 - '2.5'
+- '3.0
```
ekohl, line 3:
```suggestion
- '3.0'
```

ekohl (commented):
In this case I wonder if we should keep 2.3 Katello 3.18 is in a weird state where it's the only version with Debian support so a migration must happen there I think.

ekohl (approved):

PR #84, opened 2021-08-11
Reduce configure_pbuilder to a single script

ehelms:
The aim here is to reduce the number of scripts we have down to a single set and start with having a single script to configure pbuilder.


theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,33 @@ set -xe
...
+echo "echo deb http://deb.theforeman.org/ ${os} ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
+echo "echo deb http://deb.theforeman.org/ plugins ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
 
-  # Permit per-OS/project setups to add their own repos
-  if [ -e debian/${os}/${project}/hooks ]; then
-    sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
-  fi
+# Permit per-OS/project setups to add their own repos
+if [ -e debian/${os}/${project}/hooks ]; then
+  sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
```
evgeni, line 26: this is still using `pbuilderd` and after looking at it, I think I would prefer to have a `pbuilderd=/etc/pbuilder/${os}64` at the top of this file, and then use `pbuilderd` everywhere as before.

makes it also easier to get rid of the `64` at some point ;)

theforeman.org/scripts/debian/configure_pbuilder_plugin.sh
```diff
@@ -1,28 +0,0 @@
...
-    echo "echo deb http://stagingdeb.theforeman.org/ plugins ${repoowner} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
-  fi
-fi
-
-# Use tee to get around sudo/redirection problems, with /dev/null to drop the stdout
-echo "echo deb http://deb.theforeman.org/ ${os} ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
-echo "echo deb http://deb.theforeman.org/ plugins ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
-
-# Read Foreman package installation logs on failure
-echo "cat /var/log/foreman-install.log" | sudo tee /etc/pbuilder/${os}64/hooks/C10foremanlog
```
evgeni, line 25: this one is missing from the "new" version.

I *think* that's OK, but wanted to point out if that was an oversight.

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,33 @@ set -xe
 # This script sets up the PBuilder hooks to pull in our deb repositories while the package is building
 echo "--Apt Hook Setup"
 
-for pbuilderd in /etc/pbuilder/${os}*; do
-  # Cleanup any old hooks
-  sudo rm -f ${pbuilderd}/hooks/F60addforemanrepo
+# Cleanup any old hooks
+sudo rm -f /etc/pbuilder/${os}64/hooks/F60addforemanrepo
```
evgeni, line 8: that's interesting. currently, we allow different `deb_build_*` jobs to run in parallel (just not the same job), but that means they actually can and break each other if timed "correctly"

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,33 @@ set -xe
...
-  if [ -e debian/${os}/${project}/hooks ]; then
-    sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
-  fi
+# Permit per-OS/project setups to add their own repos
+if [ -e debian/${os}/${project}/hooks ]; then
+  sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
+fi
+
+# Read Foreman package installation logs on failure
+echo "cat /var/log/foreman-install.log" | sudo tee /etc/pbuilder/${os}64/hooks/C10foremanlog
```
evgeni, line 30:
```suggestion
echo "cat /var/log/foreman-install.log || true" | sudo tee /etc/pbuilder/${os}64/hooks/C10foremanlog
```

We shouldn't fail this if the log doesn't exist

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,33 @@ set -xe
...
-  echo "echo deb http://deb.theforeman.org/ plugins ${version} >> /etc/apt/sources.list" | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo > /dev/null
+# Use tee to get around sudo/redirection problems, with /dev/null to drop the stdout
+echo "echo deb http://deb.theforeman.org/ ${os} ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
+echo "echo deb http://deb.theforeman.org/ plugins ${version} >> /etc/apt/sources.list" | sudo tee -a /etc/pbuilder/${os}64/hooks/F60addforemanrepo > /dev/null
 
-  # Permit per-OS/project setups to add their own repos
-  if [ -e debian/${os}/${project}/hooks ]; then
-    sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
-  fi
+# Permit per-OS/project setups to add their own repos
```
evgeni, line 24: I think we should get rid of this, we don't have any hooks defined in git, and I think the less stuff we allow to execute from a PR, the better.

evgeni (changes requested):
see inline

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,30 @@ set -xe
...
+# Use tee to get around sudo/redirection problems, with /dev/null to drop the stdout
+echo "echo deb http://deb.theforeman.org/ ${os} ${version} >> /etc/apt/sources.list" | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo > /dev/null
+echo "echo deb http://deb.theforeman.org/ plugins ${version} >> /etc/apt/sources.list" | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo > /dev/null
 
-  # Permit per-OS/project setups to add their own repos
-  if [ -e debian/${os}/${project}/hooks ]; then
-    sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
-  fi
+# Read Foreman package installation logs on failure
+echo "cat /var/log/foreman-install.log" || true | sudo tee ${pbuilderd}/hooks/C10foremanlog
```
ekohl, line 27: Does this work? Shouldn't it make sure the hook doesn't fail? I can't image that `echo` would ever fail.
```suggestion
echo "cat /var/log/foreman-install.log || true"| sudo tee ${pbuilderd}/hooks/C10foremanlog
```

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,30 @@ set -xe
...
+
+# Make executable
+sudo chmod 0775 ${pbuilderd}/hooks/*
```
ekohl, line 30: If you're going this route, then I'd move it below `F99printrepos` so you can skip one chmod.

However, I wonder if there are any example hooks in there already we don't want to activate. Then again, it already was done in plugin so it's probably safe.

ehelms: I don't know enough to know what use case this covered but it was in a few of the scripts so I kept it.

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,30 @@ set -xe
...
 
-  # Permit per-OS/project setups to add their own repos
-  if [ -e debian/${os}/${project}/hooks ]; then
-    sed "s/%OS%/${os}/g" debian/${os}/${project}/hooks | sudo tee -a ${pbuilderd}/hooks/F60addforemanrepo
-  fi
+# Read Foreman package installation logs on failure
+echo "cat /var/log/foreman-install.log || true" | sudo tee ${pbuilderd}/hooks/C10foremanlog
+
+echo "tail -n 100 /etc/apt/sources.list /etc/apt/sources.list.d/*.list" | sudo tee ${pbuilderd}/hooks/F99printrepos
+sudo chmod 0775 ${pbuilderd}/hooks/F99printrepos
```
ekohl, line 30: This is redundant now that line 33 is follow it.
```suggestion
```

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,30 @@ set -xe
 # This script sets up the PBuilder hooks to pull in our deb repositories while the package is building
 echo "--Apt Hook Setup"
 
-for pbuilderd in /etc/pbuilder/${os}*; do
-  # Cleanup any old hooks
-  sudo rm -f ${pbuilderd}/hooks/F60addforemanrepo
+pdbuilderd=/etc/pbuilder/${os}64
```
ekohl, line 7:
```suggestion
pbuilderd=/etc/pbuilder/${os}64
```

That said, looks like it's only used as `${pbuilderd}/hooks`. Perhaps `$hookdir` is a better name?

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -4,29 +4,29 @@ set -xe
 # This script sets up the PBuilder hooks to pull in our deb repositories while the package is building
 echo "--Apt Hook Setup"
 
-for pbuilderd in /etc/pbuilder/${os}*; do
-  # Cleanup any old hooks
-  sudo rm -f ${pbuilderd}/hooks/F60addforemanrepo
+pbuilder_hooksdir=/etc/pbuilder/${os}64/hooks
 
+# Cleanup any old hooks
+sudo rm -f ${pbuilder_hooksdir}/hooks/F60addforemanrepo
```
ekohl, line 10: (repeated a few times)
```suggestion
sudo rm -f ${pbuilder_hooksdir}/60addforemanrepo
```

evgeni (approved):

ekohl (approved):

PR #85, opened 2021-08-11
Refactor out setup sources script for debian and add single execute s…

ehelms:
…cript

This change pulls the common execution of pbduiler into a stand-alone
script and creates focused setup sources scripts per use case.


ehelms:
Rebased and the last re-factor needed to then start turning into pipelines

theforeman.org/scripts/debian/setup_sources_core.sh
```diff
@@ -0,0 +1,84 @@
...
+  unzip archive.zip
+  mv archive/pkg/*bz2 ${project}_${VERSION}.orig.tar.bz2
+
+  # Set this in case we need it
+  LAST_COMMIT=`curl "${json_url}" | /usr/local/bin/JSON.sh -b | egrep '"lastBuiltRevision","SHA1"' | awk '{print $NF}' | tr -d \" | head -n1`
+else
+  VERSION=`echo ${VERSION} | tr '~rc' '-rc'`
+  # Download sources
+  wget http://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2
+  wget http://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2.sig
```
ekohl, line 38: While we're at it, can we change this to
```suggestion
  wget https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2 https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2.sig
```

ekohl (approved):
I haven't looked too closely, but this generally looks good.

theforeman.org/scripts/debian/setup_sources_plugin.sh
```diff
@@ -0,0 +1,65 @@
...
+    tar -x -C "${project}" -f "${project}_${VERSION}.orig.tar.gz"
+    cd ${project}
+  else
+    echo "Unsupported build type: ${BUILD_TYPE}"
+    exit 1
+  fi
+fi
+
+# Add changelog entry if this is a git/nightly build
+if [[ $gitrelease == true ]] || [[ -z $pr_number ]]; then
```
evgeni, line 48:
```suggestion
if [[ $gitrelease == true ]] || [[ -n $pr_number ]]; then
```

the original check was `[ x$pr_number != x ]` -- `pr_number` is not empty :)

theforeman.org/scripts/debian/setup_sources_core.sh
```diff
@@ -0,0 +1,83 @@
...
+  wget https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2 https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2.sig
+
+  # Verify with packaging key - commented until we can handle multiple keys
+#  tmp_keyring="./1AA043B8-keyring.gpg"
+#  gpg --no-default-keyring --keyserver keys.gnupg.net --keyring $tmp_keyring --recv-keys 1AA043B8
+#  if gpg --no-default-keyring --keyring $tmp_keyring ${project}-${VERSION}.tar.bz2.sig 2>&1 | grep -q "gpg: Good signature from \"Foreman Automatic Signing Key (2014) <packages@theforeman.org>\"" ; then
+#    true # ok
+#  else
+#    exit 2
+#  fi
```
evgeni, line 46:
```suggestion
```

unused code can go for a walk

theforeman.org/scripts/debian/setup_sources_core.sh
```diff
@@ -0,0 +1,83 @@
...
+if [[ -d ${project}-${VERSION}-develop ]] ; then
+	mv ${project}-${VERSION}-develop ${project}-${VERSION}
+fi
+
+# Bring in the debian packaging files
+cp -r ../${project} ./${project}-${VERSION}/debian
+cd ${project}-${VERSION}
+
+# Add changelog entry if this is a git/nightly build
+if [ x$gitrelease = xtrue ] || [ x$pr_number != x ]; then
```
evgeni, line 64:
```suggestion
if [[ $gitrelease == true ]] || [[ -n $pr_number ]]; then
```

theforeman.org/scripts/debian/execute_pbuilder.sh
```diff
@@ -2,86 +2,7 @@
...
-  cat debian/changelog.tmp >> debian/changelog
-  rm -f debian/changelog.tmp
-
-  # rename orig tarball to stop lintian complaining
-  mv ../${project}_${VERSION}.orig.tar.bz2 ../${project}_9999.orig.tar.bz2
-fi
+echo "--Execute pdebuild"
 
 # Build the package for the OS using pbuilder
 # needs sudo as pedebuild uses loop and bind mounts
```
evgeni, line 8: I can't comment on the lines below this, but I think they should be folded to `sudo pdebuild-${os}64`, we don't have non x86 builds anymore, and the else would actually break :)

evgeni (changes requested):
one small bug and a few comments.

overall :+1:

feels like we can slim down the "setup_sources" even further (especially the whole changelog generation), but I think we should leave that to another round once we've seen this one is good.

theforeman.org/scripts/debian/setup_sources_core.sh
ekohl: Can't we fix the TODO? Isn't that basically add an additional key to the keyring and grep for `gpg: Good signature from`?

evgeni: Grepping gpg output sounds awful, gpg has `--verify`.

But I think we should add that to the TODO list for the re-write, not try to fix it now, as it has been disabled for a while now.

ekohl: I took a stab at it and I think it should be something like this
```sh
GPG_KEYSERVER=pgp.mit.edu
GPG_KEYS="1AA043B8 C8B9E30B0DBCAB595E98DF5C965467189B776908"

# Download sources
wget https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2 https://downloads.theforeman.org/${project}/${project}-${VERSION}.tar.bz2.sig

# Verify with packaging key
tmp_keyring="$(mktemp)"
gpg2 --no-default-keyring --keyserver $GPG_KEYSERVER --keyring $tmp_keyring --recv-keys $GPG_KEYS
if gpg2 --no-default-keyring --keyring $tmp_keyring --verify ${project}-${VERSION}.tar.bz2.sig ; then
  rm $tmp_keyring
  true # ok
else
  rm $tmp_keyring
  exit 2
fi
```

And I did have some issues with pgp.mit.edu, but that's where we upload it:
https://github.com/theforeman/theforeman-rel-eng/blob/7d8f8c69f935c7912088dd5137520ddeb165d219/upload_gpg#L15

And yes, we should also specify the long key for the old key since short keys are vulnerable.

ehelms: As I do not want to pile on instability in this PR since it already is changing structure I have captured this as an issue against the milestone: https://github.com/theforeman/jenkins-jobs/issues/91

theforeman.org/scripts/debian/execute_pbuilder.sh
```diff
@@ -2,97 +2,11 @@
 set -xe
 
 # This script clones the git repos, cleans them up, and builds the package
```
ekohl, line 4: Not really true anymore. Given its size, let's drop it
```suggestion
```

theforeman.org/scripts/debian/setup_sources_core.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+set -xe
+
+# This script clones the git repos, cleans them up, and builds the package
```
ekohl, line 4:
```suggestion
# This script clones the git repos and cleans them up
```

evgeni (approved):
ewoud makes good docs points, going to ACK pending resolution of those :)

ekohl (approved):

PR #86, opened 2021-08-11
Refactor to a single stage_packages script

ehelms:


theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
...
+    echo "scratch build: uploading to stagingdeb/plugins/${repoowner}"
+    export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
+    USER=freightstage
+    COMPONENT=${repoowner}
+  fi
+else
+  echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
+  export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
+  USER=freightstage
+  HOSTS="web01.osuosl"
```
evgeni, line 36:
```suggestion
```

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
 # * the web node must have the freight class applied
 
 # Find the deps in the build-dir from the previous step
-DEB_PATH=./debian/${os}/build-${project}
+DEB_PATH=$(find ./ -name build-${project})
 
 # Upload all builds to stagingdeb for testing
-echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
-export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
-USER=freightstage
```
evgeni, line 15: I'd keep the two lines here, and only specialcase the "freight" case when uploading plugin builds.

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
 # * the web node must have the freight class applied
 
 # Find the deps in the build-dir from the previous step
-DEB_PATH=./debian/${os}/build-${project}
+DEB_PATH=$(find ./ -name build-${project})
```
ekohl, line 10: Is there no risk that it finds multiple projects now? Do we properly clean up? Also, can't we solve this with a simple glob? If anything, we can add a min and max depth so it doesn't do a very slow and fully recursive search. If there are a lot of files this can be very slow.

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
...
 HOSTS="web01.osuosl"
-COMPONENT=${repoowner}-${version}
+
+if [[ "$DEB_PATH" == *"plugins"* ]];then
+  # Don't upload scratch builds to deb.tf.o
+  if [ x$repoowner = xtheforeman ] && [ x$pr_number = x ]; then
+    echo "Built from main repo, uploading to deb/plugins/main"
+    export RSYNC_RSH="ssh -i /home/jenkins/workspace/deb_key/rsync_freight_key"
+    USER=freight
+    if [ x$repo = xdevelop ]; then
```
ekohl, line 21: If you're going for bash, might as well do it here
```suggestion
    if [[ $repo == develop ]]; then
```

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
...
+    fi
+  else
+    echo "scratch build: uploading to stagingdeb/plugins/${repoowner}"
+    export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
+    USER=freightstage
+    COMPONENT=${repoowner}
+  fi
+else
+  echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
+  export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
```
ekohl, line 34: Given the repetition of this, can't we set this outside of the loop (around line 13)?

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,44 @@ set -xe
...
+  HOSTS="web01.osuosl"
+  COMPONENT=${repoowner}-${version}
+fi
 
 for HOST in $HOSTS; do
   # The path is important, as freight_rsync (which is run on the web node for incoming
   # transfers) will parse the path to figure out the repo to send debs to.
-  TARGET_PATH="${USER}@${HOST}.theforeman.org:rsync_cache/${os}/${COMPONENT}/"
+  if [[ "$DEB_PATH" == *"plugins"* ]];then
+    TARGET_PATH="${USER}@${HOST}.theforeman.org:rsync_cache/plugins/${COMPONENT}/"
```
ekohl, line 44: While we're at it, can we make `$HOST` the FQDN and simply use `$HOST` here? I feel like that's clearer for everyone and doesn't limit us to `theforeman.org`.

theforeman.org/scripts/debian/stage_packages.sh
ehelms: I do not follow what the suggestion really is here. I can say that I think the risks are low as this is performed within a git checkout, and that checkout will be unique to the job run.

theforeman.org/scripts/debian/stage_packages.sh
ehelms: See if my change matches what you were aiming for.

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,41 @@ set -xe
...
 # Upload all builds to stagingdeb for testing
-echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
+HOSTS="web01.osuosl.theforeman.org"
 export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
 USER=freightstage
-HOSTS="web01.osuosl"
-COMPONENT=${repoowner}-${version}
+
+if [[ "$DEB_PATH" == *"plugins"* ]];then
+  # Don't upload scratch builds to deb.tf.o
```
evgeni, line 18: I think this comment is (now) misleading, as this hunk actually *does* upload to deb.tfm.o.

Maybe something like this: Plugins aren't staged. Builds from the main (theforeman) repo are uploaded directly to deb.tfm.o. Only user (PR) builds are uploaded to stagingdeb.tfm.o.

theforeman.org/scripts/debian/stage_packages.sh
evgeni: I think there is no risk in different *projects*, as the find has the project in the search criteria. But different OS'es *could* clash (I think Jenkins *does* different checkouts, but we can be safe regardless).

The old logic was:
* core: `DEB_PATH=./debian/${os}/build-${project}`
* plugins: ` DEB_PATH=./plugins/build-${project}`
* deps: `DEB_PATH=./dependencies/${os}/build-${project}`

And the problem you're trying to solve is the fact that plugins doesn't have a `${os}` subdir?

The a tad verbose, but safe solution would be:
```bash
for possible_path in ./debian/${os}/build-${project} ./plugins/build-${project} ./dependencies/${os}/build-${project}; do
  if [[ -d "${possible_path}" ]]; then
    DEB_PATH="${possible_path}"
  fi
done
```

The less vebose would be `find . -mindepth 2 -maxdepth 3 -type d -name build-${project}`

theforeman.org/scripts/debian/stage_packages.sh
evgeni: Yes, absolutely!

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,41 @@ set -xe
...
+  if [ x$repoowner = xtheforeman ] && [ x$pr_number = x ]; then
+    echo "Built from main repo, uploading to deb/plugins/main"
+    export RSYNC_RSH="ssh -i /home/jenkins/workspace/deb_key/rsync_freight_key"
```
ekohl, line 21: This is already set
```suggestion
```

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,41 @@ set -xe
...
-echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
+HOSTS="web01.osuosl.theforeman.org"
 export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
 USER=freightstage
-HOSTS="web01.osuosl"
-COMPONENT=${repoowner}-${version}
+
+if [[ "$DEB_PATH" == *"plugins"* ]];then
+  # Don't upload scratch builds to deb.tf.o
+  if [ x$repoowner = xtheforeman ] && [ x$pr_number = x ]; then
```
ekohl, line 19: We can use bash
```suggestion
  if [[ $repoowner == theforeman ]] && [[ -z $pr_number ]]; then
```

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,41 @@ set -xe
...
+fi
 
 for HOST in $HOSTS; do
   # The path is important, as freight_rsync (which is run on the web node for incoming
   # transfers) will parse the path to figure out the repo to send debs to.
-  TARGET_PATH="${USER}@${HOST}.theforeman.org:rsync_cache/${os}/${COMPONENT}/"
+  if [[ "$DEB_PATH" == *"plugins"* ]];then
+    TARGET_PATH="${USER}@${HOST}:rsync_cache/plugins/${COMPONENT}/"
+  else
+    TARGET_PATH="${USER}@${HOST}:rsync_cache/${os}/${COMPONENT}/"
```
ekohl, line 43: For what it's worth, in Debian you typically have a `suite`. This suite is normally the OS release (buster, focal, etc). However, for our plugins we don't differentiate and only have `plugins`. So in scripts you can make this clear by using `$SUITE`. I would do this around line 10 and reuse it around line 17 (`if [[ "$DEB_PATH" == *"plugins"* ]];then`)

theforeman.org/scripts/debian/stage_packages.sh
evgeni: No, it's not. It's set to `/home/jenkins/workspace/staging_key/rsync_freightstage_key`, and that key can't write to `freight@`.

ekohl: Ah my bad. I'm addressing my caffeine levels right now.

theforeman.org/scripts/debian/stage_packages.sh
ehelms: Do you happen to know *why* we don't differentiate for plugins but do for everything else?

ekohl: To save time. Most plugins aren't really OS specific: they contain gems, but `bundle install` runs on the target host. I think we're counting on plugins not containing any binaries that may be linked to system libraries. If they do, they belong in the dependencies part which is OS specific.

By doing this, we only need a single build for all plugin repos instead of n builds.

ehelms: So in theory, we could save time by ensuring that the only packages in `dependencies` are ones with binary components? For example, could https://github.com/theforeman/foreman-packaging/tree/deb/develop/dependencies/bionic/kafo_parsers be moved to plugins? Could all the hammer plugins?

evgeni: Technically? Yeah. But those don't take that long to build anyways. So your saving won't be much.

ekohl: It would also be confusing if Hammer is in main for RPMs while it's in plugins for Debian. As for kafo_parsers, then you'd have the Foreman repo depend on plugins to build. Essentially the layering is (in order):
* dependencies
* debian
* plugins

When we start building packages for a new distro version, we start with dependencies, then move on to debian and only then plugins. If you need to be aware that you first need to build some packages from plugins, then it becomes harder.

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,46 @@ set -xe
...
 export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
 USER=freightstage
-HOSTS="web01.osuosl"
-COMPONENT=${repoowner}-${version}
+SUITE=${os}
+
+if [[ "$DEB_PATH" == *"plugins"* ]];then
+  SUITE="plugins"
+
+  # Plugins aren't staged.
```
ehelms, line 25: @evgeni Why are plugins not staged?

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,46 @@ set -xe
...
+done
 
 # Upload all builds to stagingdeb for testing
-echo "scratch build: uploading to stagingdeb/${os}/${repoowner}-${version}"
+HOSTS="web01.osuosl.theforeman.org"
 export RSYNC_RSH="ssh -i /home/jenkins/workspace/staging_key/rsync_freightstage_key"
 USER=freightstage
-HOSTS="web01.osuosl"
-COMPONENT=${repoowner}-${version}
+SUITE=${os}
```
ehelms, line 20: @ekohl I am not sure if this is what you intended by using `SUITE`

ekohl: Yes, this is exactly what I meant.

theforeman.org/scripts/debian/stage_packages.sh
evgeni: For the same reason they aren't in RPM land? We don't want one plugin to hold all the plugins. (Plus we don't have gating tests for most plugins anyways)

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -7,19 +7,46 @@ set -xe
 # * the web node must have the freight class applied
 
 # Find the deps in the build-dir from the previous step
-DEB_PATH=./debian/${os}/build-${project}
+for possible_path in ./debian/${os}/build-${project} ./plugins/build-${project} ./dependencies/${os}/build-${project}; do
+  if [[ -d "${possible_path}" ]]; then
+    DEB_PATH="${possible_path}"
+  fi
+done
```
ekohl, line 14: Just for safety, could we add some code after this loop? Something like 
```suggestion
for possible_path in ./debian/${os}/build-${project} ./plugins/build-${project} ./dependencies/${os}/build-${project}; do
  if [[ -d "${possible_path}" ]]; then
    DEB_PATH="${possible_path}"
  fi
done

if [[ -z $DEB_PATH ]] ; then
  echo "No build-dir path found"
  exit 1
fi
```

I also wonder about the chance where it found something that wasn't intended, but I think that chance is small since projects only live in one directory and I don't recall us ever moving something. Famous last words, but I think it's safe to ignore that case for now.

evgeni: There is a `set -e` at the top of this script, so I'd expect it to abort if it tries to access an empty `DEB_PATH`, so the safeguard is already there?

evgeni: Should we just `set -u`?

theforeman.org/scripts/debian/stage_packages.sh
ehelms: Technically, they are staged in RPM land, as we mash the repository into a staging repository on Koji, then we perform repoclosure and then we promote the repository to production location.

theforeman.org/scripts/debian/stage_packages.sh
ekohl: Could be, but I must admit I don't have a lot of experience with that. If it'd happen more often a clear error message can be helpful but I think we'd be fine with it given the chance is small.

theforeman.org/scripts/debian/stage_packages.sh
evgeni: As we don't perform any repoclosure on Debian… ;)

ehelms: I get that part, I was more wanting to understand if it was a technical reason or a process reason. Different processes make automating, and understanding harder. What would be the technical downside of staging them first? Time and space?

evgeni: Space is probably negligible. Time… Well, it takes 10-ish minutes to publish Debian, so we'd add that time somewhere (unless we add the unstaging to the normal Foreman pipe).
So, what would unstage them? A normal Foreman pipeline? A dedicated pipeline that does effectively nothing?

ekohl (approved):

theforeman.org/scripts/debian/stage_packages.sh
```diff
@@ -1,25 +1,57 @@
...
+
+if [[ -z $DEB_PATH ]] ; then
```
evgeni, line 16: the script is now `-u` so this will explode and the `echo` won't trigger.
```suggestion
if [[ -z ${DEB_PATH:-} ]] ; then
```

evgeni (approved):
shell-nitpick, but the current version does what it should, so ACK

PR #90, opened 2021-08-25
Replace JSON.sh with jq

ekohl:
We currently ship JSON.sh in our infra module. This depends on a standard tool instead.

Depends on https://github.com/theforeman/foreman-infra/pull/1638


ekohl:
I should note that I manually tested the JQ commands, but I'm not sure I caught everything.

PR #96, opened 2021-08-29
Add pipeline for Debian PR testing and Debian package release

ehelms:


ehelms:
Contains https://github.com/theforeman/jenkins-jobs/pull/97 which should be merged first to make it easier to read this diff

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,455 @@
...
+            steps.stage_packages["${os}-${project}"] = {
+                stage_debian_packages(pkg_to_build['type'], project, os, version, repoowner, pull_request)
+            }
+        }
+    }
+
+    return steps
+}
+
+def configure_pbuilder(project, os, version, repoowner) {
```
ehelms, line 153: This depends on the outcome of https://github.com/theforeman/jenkins-jobs/pull/95/files and https://github.com/theforeman/foreman-infra/pull/1642

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,455 @@
...
+
+            sh """
+              # If a last* alias was used, resolve the numeric job ID
+              job_id=\$(curl "${json_url}" | /usr/local/bin/JSON.sh -b | awk '\$1 == "[\"number\"]" { print \$2 }')
+              base_url=`curl "${json_url}" | /usr/local/bin/JSON.sh -b | awk '\$1 ~ /^\\["runs",.*,"number"\\]/ && \$2 == '\$job_id' {getline; print \$2; exit}' | tr -d \"`
+
+              if [ x\$base_url = x ] ; then
+                base_url=`curl "${json_url}" | /usr/local/bin/JSON.sh -b | egrep '\\["url"\\]' | awk '{print \$NF}' | tr -d \"`
+              fi
+              url="${base_url}/artifact/*zip*/archive.zip"
```
evgeni, line 241: This can be simplified if we merge https://github.com/theforeman/jenkins-jobs/pull/90 first :)

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -0,0 +1,455 @@
...
+
+              wget \$url
+              unzip archive.zip
+              mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
+
+              # Set this in case we need it
+            """
+
+            last_commit = sh(returnStdout: true, script: "curl \"${json_url}\" | /usr/local/bin/JSON.sh -b | egrep '\"lastBuiltRevision\",\"SHA1\"' | awk '{print \$NF}' | tr -d \" | head -n1").trim()
```
evgeni, line 250: same, https://github.com/theforeman/jenkins-jobs/pull/90 makes it easier

evgeni:
#97 is merged

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,348 @@ def repoclosures(repo, releases, version) {
...
+      cat debian/changelog.tmp >> debian/changelog
+      rm -f debian/changelog.tmp
+    """
+}
+
+def execute_pbuilder(build_dir, os) {
+    dir(build_dir) {
+        sh """
+            # Build the package for the OS using pbuilder
+            # needs sudo as pedebuild uses loop and bind mounts
```
ekohl, line 405: Should this be `pdebuild` instead of `pdebuild`?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,302 @@ def repoclosures(repo, releases, version) {
...
+
+    dir(build_dir) {
+        if (version =~ 'develop') {
+            def url_base = 'https://ci.theforeman.org'
+            def job_name = nightly_jenkins_job
+            def job_id = nightly_jenkins_job_id
+            def job_url = "${url_base}/job/${job_name}/${job_id}"
+
+            sh """
+              wget "${job_url}/artifact/*zip*/archive.zip"
```
ekohl, line 197: I'm wondering if we could use a native method like [copyArtifacts](https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.step.StepContext.copyArtifacts).

theforeman.org/pipelines/release/foreman_packaging_deb_release.groovy
```diff
@@ -0,0 +1,94 @@
...
+                script {
+                    parallel build_steps.executes
+                }
+            }
+        }
+
+        stage('Stage packages') {
+            steps {
+                script {
+                    parallel build_steps.stage_packages
```
ekohl, line 73: I don't think this should be done in parallel. We rsync them over and then rebuild the archive. That's quite an expensive operation and if you can do it just once for multiple packages that's much cheaper.

ehelms: We already do this in parallel as I understand things. The original stage packages script is run per project-os combination. So a job like https://ci.theforeman.org/job/packaging_build_deb_dependency/559/console would run that script three times which could all happen at the same time? 

What rebuilds the archive? the rsync trigger that?

ekohl: On the server side there's a script that runs when rsync runs:
https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/freight/templates/rsync.erb

As you can see, it runs the original command and then calls
```erb
  find <%= @home %>/rsync_cache/$DISTRO/$REPO -iname '*.deb' -exec freight-add -v -c <%= @home %>/freight.conf {} apt/$DISTRO/$REPO \;
  # Publish the debs
  freight-cache -c <%= @home %>/freight.conf -v apt/$DISTRO
  # Cleanup - no need to keep the debs
  find <%= @home %>/rsync_cache/$DISTRO/$REPO -iname '*.deb' -delete || true
```

ehelms: So what matters is `distro` and `repo`, so we could run this in parallel per OS and repository if we collected all those possible combinations up and constructed the run based on that?

ehelms: Sorry, what I meant to ask was, is there a model where we could initiate this script directly for distro/repo pairings based on what was built given that the rsync is based upon the actual projects that are built?

ekohl: Yes, we essentially need to gather all changed debs. Then make 2 lists:
* Those to staging (core + dependency)
* Those to release (plugin)

Then roughly `if (to_staging.any?) { rsync_to_staging(to_staging) }` and the same with `to_release`.

You're right that this is an optimization we didn't do before, but can do now.

ehelms: Is there anything that would prevent us from moving to a model where all of or parts of https://github.com/theforeman/foreman-infra/blob/master/puppet/modules/freight/templates/rsync.erb are initiated explicitly rather than implicitly due to an rsync? I am thinking that would allow us to have cleaner optimized code if the archive building is the expensive operation.

ehelms:
There is a bit to digest here, and where it made sense I have tried to break things out from shell into native groovy methods of code handling. The code is still a bit complex due to:

 * source generation requires different methods based on a few criteria
 * plugins are handled special in almost every case making for a lot of variation
    - their directory structure
    - that they are not staged unless its a pull request
  
I wonder if there are further optimizations we could make in the process to make it smoother to understand the branches. For example, choosing to stage plugins rather than release them directly to reduce those code branches. 

These are all things that could be done before hand or after this is merged. Given this provides various benefits to get merged (e.g. an auto release job).

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+        if (pull_request) {
+            last_commit = sh(reutrnStdout: true, script: "git rev-list HEAD|/usr/bin/head -n 1").trim()
+            add_debian_changelog(os, package_version, repoowner, last_commit)
+        }
+    }
+
+    return "${build_dir}/${package_dir}"
+}
+
+def add_debian_changelog(os, package_version, repoowner, last_commit) {
```
ehelms, line 368: Attempting to make this change here first to reduce the diff https://github.com/theforeman/jenkins-jobs/pull/100

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+
+            if [[ \$BUILD_TYPE == "python" ]]; then
+              pip download --no-deps --no-binary=:all: "${project}==${package_version}"
+              mkdir "${project}-${package_version}"
+              tar -x -C "${project}-${package_version}" --strip-components=1 -f "${project}-${package_version}.tar.gz"
+              mv "${project}-${package_version}.tar.gz" "${project}_${package_version}.orig.tar.gz"
+              cp -r ../${project} "${project}-${package_version}/debian"
+            elif [[ \$BUILD_TYPE == "gem" ]]; then
+              gem fetch ${project} -v "=${package_version}"
+              ../../gem2deb ${project}-${package_version}.gem --debian-subdir ../${project} --only-source-dir
```
ehelms, line 291: I am wondering if we should pull `gem2deb` to a stand-alone repository (or foreman-infra) and deploy it to the Jenkins nodes?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+                BUILD_TYPE=gem
+                # Import variables from the project, allowing it to override behaviour
+                if [ -e ../${project}/build_vars.sh ]; then
+                  . ../${project}/build_vars.sh
+                fi
+
+                if [ "\$BUILD_TYPE" = "gem" ]; then
+                    cp -r ../${project} ./
+                    cd ${project}
+                    ../../download_gems
```
ehelms, line 341: Anybody know why for plugins that are gems we run this download_gems script and for other gems (like dependencies) we use `gem fetch` and `gem2deb` ?

theforeman.org/pipelines/lib/packaging.groovy
evgeni: Maybe. The crux behind *this* `gem2deb` is: it's a patched version :/

(not saying we can make deployment of it easier, just that we need _that_ version)

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+              exit 1
+            fi
+        """
+
+        package_dir = sh(returnStdout: true, script: "find -maxdepth 1 -type d -not -name '.'").trim()
+    }
+
+    dir("${build_dir}/${package_dir}") {
+        if (pull_request) {
+            last_commit = sh(reutrnStdout: true, script: "git rev-list HEAD|/usr/bin/head -n 1").trim()
```
evgeni, line 303: we already have `archive_git_hash` in `git.groovy`, which does `git rev-parse ${ref}` (and that's equal to `rev-list and then getting the first entry)

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+
+    return "${build_dir}/${package_dir}"
+}
+
+def setup_sources_plugin(project, os, version, repoowner, pull_request = false) {
+    def package_version
+    def build_dir = deb_build_dir('plugin', project)
+    def package_dir
+
+    sh "mkdir -p ${build_dir}"
```
evgeni, line 316: doesn't `dir(…)` do that for us already?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+            package_version = sh(returnStdout: true, script: "head -n1 ../${project}/changelog|awk '{print \$2}'|sed 's/(//;s/)//'|cut -f1 -d-").trim()
+
+            sh """
+                gem fetch ${project} -v "=${package_version}"
+                ../../dependencies/gem2deb ${project}-${package_version}.gem --debian-subdir ../${project} --only-source-dir
+            """
+
+            package_dir = sh(returnStdout: true, script: "find -maxdepth 1 -type d -not -name '.'").trim()
+        } else {
+            package_version = sh(returnStdout: true, script: "head -n1 ../${project}/debian/changelog|awk '{print \$2}'|sed 's/(//;s/)//'|cut -f1 -d-|cut -d: -f2").trim()
```
evgeni, line 329: we have similar (but not identical) versions (heh) of this in various places, and I think this is the most correct one, so how about

```groovy
def package_version(changelog_path) {
    return sh(returnStdout: true, script: "head -n1 ${changelog_path} |awk '{print \$2}'|sed 's/(//;s/)//'|cut -f1 -d-|cut -d: -f2").trim()
```

and then using it everywhere

theforeman.org/pipelines/lib/packaging.groovy
evgeni: I think (and I can be totally wrong) this is due to the totally different ways of how we build/vendor stuff.
* dependencies are "real" (mostly policy compliant) Debian packages with Ruby modules built against the system Ruby, so the workflow is: use `gem fetch` to obtain the `.gem` file, call `gem2deb` on it to convert it to a Debian (source) package, build it.
* core is a big bundle (generated by `bundle package`), with some system level dependencies from above.
* plugins are *extending* this (the core) bundle, to be able to do this, we pull down the gems (of the plugin and its dependencies, using `download_gems`) and place those in the package into the same folder the core bundle is installed (and a bundler.d snippet). on installation of the package, we call `bundle update` and thus actually add those modules to the bundle.

hope this helps, and didn't confuse you further?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+      rm -f debian/changelog.tmp
+    """
+}
+
+def execute_pbuilder(build_dir, os, version) {
+    dir(build_dir) {
+        sh """
+            # Build the package for the OS using pbuilder
+            # needs sudo as pdebuild uses loop and bind mounts
+            sudo FOREMAN_VERSION=${version} pdebuild-${os}64
```
evgeni, line 393:
```suggestion
            FOREMAN_VERSION=${version} pdebuild-${os}64
```

https://github.com/theforeman/jenkins-jobs/pull/99 changed that.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,319 @@ def repoclosures(repo, releases, version) {
...
+
+            # Cleanup, pdebuild uses root
+            sudo chown -R jenkins:jenkins $WORKSPACE
```
evgeni, line 396: I *think* this is not needed anymore after #99, but I am not 100% sure.

theforeman.org/pipelines/lib/packaging.groovy
ekohl: I think only Foreman plugins use bundler that way. Smart Proxy plugins are more policy compliant and AFAIK also use system Ruby (and system gems).

evgeni: That's of course correct and makes me no more happier ;)

theforeman.org/pipelines/lib/packaging.groovy
ehelms: Yea... thats why I thought about pulling it to some small dedicated repo since it won't change much or drop it in `puppet/slave/files` and deploy it to the nodes directly.

ehelms:
Pushed another round of updates, I think the things being discussed currently here can happen after this goes in. I am trying to balance improvements with getting a release job available to reduce overhead with debian packaging.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,309 @@ def repoclosures(repo, releases, version) {
...
+    " > debian/changelog
+
+      cat debian/changelog.tmp >> debian/changelog
+      rm -f debian/changelog.tmp
+    """
+}
+
+def execute_pbuilder(build_dir, os, version) {
+    dir(build_dir) {
+        sh "FOREMAN_VERSION=${version} pdebuild-${os}64"
```
ehelms, line 387: @evgeni so what should this be now?

evgeni: sudo, sorry!

ehelms:
There are a couple options we could choose to deploy this and get comfortable with it:

 * Deploy just the PR testing job alongside the current PR testing job and let them run together, wait on enabling the release job
 * Replace current PR testing job with this one and let it run for a it and then enable the release job
 * Turn on just the release job to test the new code and leave existing PR testing / jobs in place and then turn on PR testing later
 * Just do it: turn everything new on and remove the old and fix it as we go

theforeman.org/pipelines/lib/packaging.groovy
ehelms: This now contains the fix from #100

theforeman.org/pipelines/lib/git.groovy
```diff
@@ -11,8 +11,13 @@ def ghprb_git_checkout() {
...
 }
 
+def git_hash(ref = 'HEAD') {
+    return sh(script: "git rev-parse ${ref}", returnStdout: true).trim()
+}
+
 def archive_git_hash(ref = 'HEAD') {
-    def hash = sh(script: "git rev-parse ${ref} | tee commit", returnStdout: true).trim()
+    def hash = git_hash(ref)
+    writeFile('commit', dash)
```
evgeni, line 20:
```suggestion
    writeFile('commit', hash)
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -34,6 +34,51 @@ def find_changed_packages(diff_range) {
     return changed_packages.split()
 }
 
+def find_changed_debs(diff_range) {
+    def dependencies = []
+    def plugins = []
+    def cores = []
```
evgeni, line 40: what's the idea behind noting them down individually, and then returning all together? ain't it "easier" (to the eyes) to `def changed_debs = []`, then in each `for …` do `changed_debs.add(…)` and later just `return changed_debs`?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,310 @@ def repoclosures(repo, releases, version) {
...
+
+    for(int i = 0; i < packages_to_build.size(); i++) {
+        def index = i
+        def pkg_to_build = packages_to_build[index]
+        def project = pkg_to_build['name']
+
+        echo "${pkg_to_build}"
+
+        for(int j = 0; j < pkg_to_build['operating_systems'].size(); j++) {
+            def os_index = j
```
evgeni, line 133: this is never used again?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,310 @@ def repoclosures(repo, releases, version) {
...
+            }
+
+            steps.executes["${os}-${project}"] = {
+                execute_pbuilder(execute_pbuilder_dir, os, version)
+            }
+
+            if (type == 'plugin' && !pull_request) {
+                suite = 'plugins'
+                component = version =~ 'develop' ? 'nightly' : version
+                release_type = 'release'
```
evgeni, line 150:
```suggestion
                release_type = 'release'
                echo "plugin build without PR: uploading directly to deb/${suite}/${component}"
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,310 @@ def repoclosures(repo, releases, version) {
...
+                suite = os
+                component = "${repoowner}-${version}"
+                release_type = 'stage'
+                echo "scratch build: uploading to stagingdeb/${suite}/${component}"
+            }
+
+            if (!repos.containsKey("${suite}-${component}")) {
+                repos["${suite}-${component}"] = [suite: suite, component: component, deb_paths: [], type: release_type]
+            }
+            repos["${suite}-${component}"].deb_paths.add("${deb_path}/*deb}")
```
evgeni, line 166:
```suggestion
            repos["${suite}-${component}"].deb_paths.add("${deb_path}/*deb")
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,310 @@ def repoclosures(repo, releases, version) {
...
+
+def setup_sources_core(os, project, version, repoowner, pull_request = false) {
+    def package_version = debian_package_version("debian/${os}/${project}/changelog")
+    def build_dir = deb_build_dir('core', project, os)
+
+    package_version = package_version.replaceAll('~rc', '-rc')
+    last_commit = package_version
+
+    dir(build_dir) {
+        if (version =~ 'develop') {
```
evgeni, line 225: why `=~` and not `=`? also this should be `nightly`, not `develop`, for the same reasons as in https://github.com/theforeman/jenkins-jobs/pull/96/files#r700835168

theforeman.org/pipelines/release/foreman_packaging_deb_release.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+                    if (fileExists('commit')) {
+
+                        commit = readFile(file: 'commit').trim()
+                        packages_to_build = find_changed_debs("${commit}..HEAD")
+                        build_steps = build_deb_package_steps(packages_to_build, foreman_version, 'theforeman', false)
+
+                    } else {
+
+                        build_steps
```
evgeni, line 47:
```suggestion
                        build_steps = []
```

evgeni: or whatever is the correct syntax, to make the other steps be noop. probably `build_steps = ['setup_sources': [], 'executes': [], 'rsync_packages': []]` or so?

theforeman.org/pipelines/test/deb_packaging.groovy
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        stage('Find Packages to Build') {
+            steps {
+
+                script {
+                    def version = ghprbTargetBranch == 'deb/develop' ? 'nightly' : ghprbTargetBranch.replace('deb/', '')
+                    packages_to_build = find_changed_debs("origin/${ghprbTargetBranch}")
+
+                    build_steps = build_deb_package_steps(packages_to_build, version, ghprbPullAuthorLogin, binding.hasVariable('ghprbPullId'))
```
evgeni, line 31: is there a case where the *test* pipeline is triggered without a PR? you do rely on `ghprbPullAuthorLogin`, so IMHO we can just assume that `pull_request=true`.

theforeman.org/scripts/debian/build_project.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
evgeni, line 1: what is this file for?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +108,310 @@ def repoclosures(repo, releases, version) {
...
+            sh """
+              # Download sources
+              wget https://downloads.theforeman.org/${project}/${project}-${package_version}.tar.bz2 https://downloads.theforeman.org/${project}/${project}-${package_version}.tar.bz2.sig
+              mv ${project}-${package_version}.tar.bz2 ${project}_${package_version}.orig.tar.bz2
+            """
+        }
+
+        sh """
+            # Unpack
+            tar xvjf ${project}_${package_version}.orig.tar.bz2
```
evgeni, line 250:
```suggestion
            tar xf ${project}_${package_version}.orig.tar.bz2
```

* `j` is not needed as `tar` can figure out the compression on its own
* `v` makes the build log so noisy, that I'd prefer if we would drop it

theforeman.org/pipelines/lib/git.groovy
ehelms: huh, wth was I thinking

theforeman.org/pipelines/lib/packaging.groovy
ehelms: Holdovers from trying to translate shell scripts to groovy. Thanks for the catch.

theforeman.org/pipelines/test/deb_packaging.groovy
ehelms: Yes, that is true now. When I wrote this I was still potentially re-using this for the release pipeline.

theforeman.org/scripts/debian/build_project.sh
ehelms: Absolutely nothing. Was accidental add from when I thought about trying to create a consolidated build script that combined all the ways to build source and then realized how crazy it is.

theforeman.org/pipelines/lib/git.groovy
evgeni: dash is the default `/bin/sh` on Debian, so idk?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -34,6 +34,49 @@ def find_changed_packages(diff_range) {
...
 
+def find_changed_debs(diff_range) {
+    def changed_debs = []
+
+    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*/changelog').split()
+    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*/changelog').split()
+    def changed_core = find_added_or_changed_files(diff_range, 'debian/*/*/changelog').split()
+
+    for(core in changed_core) {
+        core_split = core.split('/')
```
evgeni, line 45: I think you can use something link this:

```suggestion
        (folder, os, project) = core.split('/')
```

and then use the better names below.

```console
groovy:000> s = "debian/focal/foreman/changelog"
===> debian/focal/foreman/changelog
groovy:000> (type, os, project) = s.split('/')
===> [Ljava.lang.String;@dc7df28
groovy:000> type
===> debian
groovy:000> project
===> foreman
groovy:000> os
===> focal
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +106,309 @@ def repoclosures(repo, releases, version) {
...
+
+    def repos = [:]
+
+    for(int i = 0; i < packages_to_build.size(); i++) {
+        def pkg_to_build = packages_to_build[i]
+        def project = pkg_to_build['name']
+
+        echo "${pkg_to_build}"
+
+        for(int j = 0; j < pkg_to_build['operating_systems'].size(); j++) {
```
evgeni, line 129: looking at `find_changed_debs`, I don't see how `operating_systems` can be longer than 1, as you add a new entry for every changed changelog, instead of updating a possibly existing one.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -34,6 +34,49 @@ def find_changed_packages(diff_range) {
...
+    }
+
+    for(dependency in changed_dependencies) {
+        dependency_split = dependency.split('/')
+
+        changed_debs.add([
+            type: 'dependency',
+            name: dependency_split[2],
+            path: "${dependency_split[0]}/${dependency_split[1]}/${dependency_split[2]}",
+            operating_systems: dependency_split[1]
```
evgeni, line 62:
```suggestion
            operating_systems: [dependency_split[1]]
```

otherwise this will be a string, not an array?

theforeman.org/pipelines/lib/packaging.groovy
ehelms: Good catch, that logic changed so many times I missed this simplification

evgeni:
There is a syntax error somewhere, as all release related pipelines produce the following error:

```
{'error': "expecting ')', found '=' @ line 446, column 21."}
output/foreman-2.3-release-pipeline/config.xml: NOT OK
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -63,3 +106,306 @@ def repoclosures(repo, releases, version) {
...
+
+def setup_sources_core(os, project, version, repoowner, pull_request = false) {
+    def package_version = debian_package_version("debian/${os}/${project}/changelog")
+    def build_dir = deb_build_dir('core', project, os)
+
+    package_version = package_version.replaceAll('~rc', '-rc')
+    last_commit = package_version
+
+    dir(build_dir) {
+        if (version = 'nightly') {
```
evgeni, line 219:
```suggestion
        if (version == 'nightly') {
```

makes the linter so much happier ;)

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -34,6 +34,49 @@ def find_changed_packages(diff_range) {
...
 
+def find_changed_debs(diff_range) {
+    def changed_debs = []
+
+    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*/changelog').split()
+    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*/changelog').split()
+    def changed_core = find_added_or_changed_files(diff_range, 'debian/*/*/changelog').split()
+
+    for(core in changed_core) {
+        (type, os, project) = core.split('/')
```
evgeni, line 45: in https://github.com/theforeman/jenkins-jobs/pull/96#discussion_r701655615, I used `(folder, os, project)`, as it felt like using `type` here will welcome someone to refactor the line 3 down (`type: 'core'`) to read `type: type` and that'd break things.

not blocking on this, but I think not calling this thing "type" is cleaner.

ehelms: type type type

evgeni (approved):
there are still too many shell blocks, but I think we can tackle those one by one.

ehelms:
This now disables the current PR test and adds but keeps disabled the new release job. I am aiming to merge this on 9/7 to actively test Debian PR. Hopefully there will be a few open to try this out on.

PR #98, opened 2021-08-30
hooks cleanup

evgeni:


theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -5,28 +5,14 @@ set -xe
 echo "--Apt Hook Setup"
 
 pbuilder_hooksdir=/etc/pbuilder/${os}64/hooks
+local_hooksdir=$(PWD)/hooks-${os}-${version}-${BUILD_ID}
 
-# Cleanup any old hooks
-sudo rm -f ${pbuilder_hooksdir}/F60addforemanrepo
+cp --recursive --preserve=mode ${pbuilder_hooksdir} ${local_hooksdir}
```
evgeni, line 10: `pbuilder` doesn't accept multiple `--hookdir`s (well, it does, silently only using the last one), so we need to compose a hookdir with all the static hooks we deploy via Puppet *plus* the one "dynamic" we generate in here.

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -5,28 +5,14 @@ set -xe
...
 
-if [[ $repoowner != theforeman ]]; then
-  # Add a stagingdeb repo to the sources too, if it exists
-  if wget -O/dev/null -q http://stagingdeb.theforeman.org/dists/${os}/${repoowner}-${version} ; then
-    echo "echo deb http://stagingdeb.theforeman.org/ ${os} ${repoowner}-${version} >> /etc/apt/sources.list" | sudo tee -a ${pbuilder_hooksdir}/F60addforemanrepo > /dev/null
-  fi
-  if wget -O/dev/null -q http://stagingdeb.theforeman.org/dists/plugins/${repoowner} ; then
-    echo "echo deb http://stagingdeb.theforeman.org/ plugins ${repoowner} >> /etc/apt/sources.list" | sudo tee -a ${pbuilder_hooksdir}/F60addforemanrepo > /dev/null
-  fi
-fi
```
evgeni, line 20: as discussed, this is gone.

one thing that I wonder tho, this didn't add the staging repo for `theforeman` builds, which means we couldn't ever build against updated dependencies without a pipeline (which could possibly fail?!) that sounds wrong to me.

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -5,28 +5,14 @@ set -xe
 echo "--Apt Hook Setup"
 
 pbuilder_hooksdir=/etc/pbuilder/${os}64/hooks
```
evgeni, line 7: this directory doesn't contain *anything* specific to the `os` anymore, so we can move everything to `/etc/pbuilder/hooks` later.

theforeman.org/scripts/debian/configure_pbuilder.sh
```diff
@@ -5,28 +5,14 @@ set -xe
...
+local_hooksdir=$(PWD)/hooks-${os}-${version}-${BUILD_ID}
```
evgeni, line 8: or should this be `$(WORKSPACE)/hooks…`? I never know.

theforeman.org/scripts/debian/configure_pbuilder.sh
ekohl: Perhaps that's a good comment in the script itself?

theforeman.org/scripts/debian/configure_pbuilder.sh
ekohl: Indeed, then it would be different from Koji where we do update the buildroots with newly built packages. We should continue to include staging. I'd also expect that it is especially needed for `theforeman`. Perhaps that's a regression we haven't dealt with yet. Maybe we haven't seen it yet since we don't have a lot of build dependencies on our own packages and need that even less.

evgeni: indeed. this is a regression in 08820f20a781886c54bc50bf035ba029de25e995

`configure_pbuilder` used to add the `$repoowner` repo  unconditionally, and only `configure_pbuilder_plugin` had the `if`

evgeni: added back in simplified form

theforeman.org/scripts/debian/configure_pbuilder.sh
ehelms: I think `$WORKSPACE` is more predictable and helps guarantee the scope of this setup.

theforeman.org/scripts/debian/execute_pbuilder.sh
```diff
@@ -6,7 +6,7 @@ echo "--Execute pdebuild"
 
 # Build the package for the OS using pbuilder
 # needs sudo as pedebuild uses loop and bind mounts
-sudo pdebuild-${os}64
+sudo pdebuild-${os}64 --hookdir ${local_hooksdir}
```
evgeni, line 9: with https://github.com/theforeman/foreman-infra/pull/1643, this would become

```
sudo REPO_SETUP=true REPO_SUITE=${os} REPO_VERSION=${version} pdebuild-${os}64
```

technically we could even template it to not need `OS` and then default `VERSION` to `nightly`

ekohl: I like being explicit. It's only scripts that call it anyway so IMHO there's no need for human conveniences.

ekohl:
I like https://github.com/theforeman/foreman-infra/pull/1643. Could you update the PR to reflect it?

ekohl (approved):
:+1: pending the foreman-infra merge.

PR #100, opened 2021-08-30
Add consolidated add changelog script for Debian

ehelms:


ehelms:
https://github.com/theforeman/jenkins-jobs/pull/96/files#diff-c12500cef361f017ede0862bcabf9b799c8e4b9287be181a1d74203617208968R368 introduces a consolidated changelog script, to reduce the diff of that change trying to make it here.

theforeman.org/scripts/debian/add_changelog.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -xe
+
+if [[ $gitrelease == true ]] || [[ -n $pr_number ]]; then
+
+  echo "--Adding changelog entry"
+
+  PACKAGE_NAME=$(head -n1 debian/changelog|awk '{print $1}')
+  DATE=$(date -R)
+  RELEASE="9999-${VERSION}-${os}+scratchbuild+${BUILD_TIMESTAMP}"
```
evgeni, line 10: `os` is unset for the plugins case, which might generate "funny" versions here, but we could also just do `os=plugins` in `setup_sources_plugin.sh` which will accidentally produce exactly the right version string ;)

evgeni (changes requested):
tiny comment, once implemented :+1:

theforeman.org/scripts/debian/add_changelog.sh
ehelms: I think this is set right? https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/yaml/jobs/packaging_build_deb_plugin.yaml#L13

evgeni: oh, right, it is, but it's "wrong", as the plugins aren't specific to bionic.

okay, so maybe the best approach would be: for core and deps, we set `SUITE=${os}` (like we do in `stage_pacakges`), and for plugins we set `SUITE=plugins` and use `${SUITE}` not `${os}` here?

evgeni (approved):

theforeman.org/scripts/debian/add_changelog.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+set -xe
+
+if [[ $gitrelease == true ]] || [[ -n $pr_number ]]; then
+
+  echo "--Adding changelog entry"
+
+  PACKAGE_NAME=$(head -n1 debian/changelog|awk '{print $1}')
+  DATE=$(date -R)
+  RELEASE="9999-${VERSION}-${suite}+scratchbuild+${BUILD_TIMESTAMP}"
```
ekohl, line 10: Where do we set `$suite`? Is it exported somewhere or does it rely on everything being concatted?

ehelms: The latter for right now. This is handled much more cleanly in the pipeline based PR.

PR #101, opened 2021-08-31
Create an internal folder to hold all JJB jobs

ehelms:
From reading, it appeared that to enable Jenkinsfiles without access to all credentials in the system, folders need to be used. By putting credentials and jobs in the same folder, only jobs in that folder can access credentials also in said folder. While we can aim for more granularity over time my idea was the following:

 * create  an internal/JJB folder and put all jobs created via JJB in said folder
 * add all credentials to said folder and have 0 global credentials
 * put multibranch created jobs into a different folder thereby cutting off credential access

That would maintain the current state of our design while enabling to start trying out Jenkinsfiles.

### DIFF
--- a/theforeman.org/yaml/jobs/internal_folder.yaml
+++ b/theforeman.org/yaml/jobs/internal_folder.yaml
@@ -1,0 +1,4 @@
+---
+- job:
+    name: internal-folder
+    project-type: folder
--- a/theforeman.org/yaml/defaults/defaults.yaml
+++ b/theforeman.org/yaml/defaults/defaults.yaml
@@ -13,4 +13,4 @@
       - tfm-build-discarder
     wrappers:
       - timestamps
-
+    folder: internal-folder


theforeman.org/yaml/jobs/internal_folder.yaml
```diff
@@ -0,0 +1,4 @@
+---
+- job:
+  name: internal-folder
```
evgeni, line 3: jenkins_jobs.errors.JenkinsJobsException: Syntax error, for item named 'internal-folder'. Missing indent?

theforeman.org/yaml/jobs/internal_folder.yaml
```diff
@@ -0,0 +1,4 @@
...
+  project-type: folder
```
evgeni, line 4:
```suggestion
    name: internal-folder
    project-type: folder
```

ehelms:
> conceptual question: will this show up in URLs etc? as we have some tooling that relies on URLs being "stable" (like the scripts in `theforeman-rel-eng` repo)

The answers appears to be yes. I created a Test folder and then a test job and the URL for the job was:

https://ci.theforeman.org/job/Test/job/test-folder-pipe/

PR #104, opened 2021-09-07
Check for any changes to perform debian PR builds

ehelms:


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -37,9 +37,9 @@ def find_changed_packages(diff_range) {
 def find_changed_debs(diff_range) {
     def changed_debs = []
 
-    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*/changelog').split()
-    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*/changelog').split()
-    def changed_core = find_added_or_changed_files(diff_range, 'debian/*/*/changelog').split()
+    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*').split()
+    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*).split()
```
ekohl, line 41:
```suggestion
    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*').split()
```

ekohl (commented):
Is this really true? I think only a changelog entry can determine a new version. What happens if a version already exists but someone changed a README or whatever. Does it fail to build then?

ehelms:
heh, you were too quick to the draw, I was posting my thoughts here -- https://github.com/theforeman/foreman-packaging/pull/7073#issuecomment-914335446 as it triggered it

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -37,9 +37,9 @@ def find_changed_packages(diff_range) {
 def find_changed_debs(diff_range) {
     def changed_debs = []
 
-    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*/changelog').split()
-    def changed_plugins = find_added_or_changed_files(diff_range, 'plugins/*/changelog').split()
-    def changed_core = find_added_or_changed_files(diff_range, 'debian/*/*/changelog').split()
+    def changed_dependencies = find_added_or_changed_files(diff_range, 'dependencies/*/*').split()
```
evgeni, line 40: I think at this point, you can just drop the `*/*` at the end and make it `find_added_or_changed_files(diff_range, 'dependencies/*/*')`?

But hah, nope, the whole logic (below) relies on the fact that this returns ONE result per package, but now it would probably return multiple (when e.g. changelog and rules files change)

ehelms:
Closing based on thoughts here -- https://github.com/theforeman/foreman-packaging/pull/7073

PR #110, opened 2021-09-24
Add email list notification

lzap:
A proposal to create a list people could easily subscribe to:

https://community.theforeman.org/t/how-to-subscribe-to-ci-plugin-failures/25328/6

For now, I am just adding plugins develop tests, but once this is approved I will modify all places so emails are sent for everything.

### DIFF
--- a/theforeman.org/yaml/publishers/discourse.yaml
+++ b/theforeman.org/yaml/publishers/discourse.yaml
@@ -1,0 +1,5 @@
+- publisher:
+    name: discourse
+    publishers:
+    - email:
+        recipients: ci@community.theforeman.org
--- a/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
+++ b/theforeman.org/yaml/jobs/plugins/foreman_discovery.yaml
@@ -9,4 +9,7 @@
       - 18.0-stable:
           foreman_branch: 3.0-stable
     jobs:
-      - 'test_plugin_{name}_{branch}'
+      - 'test_plugin_{name}_{branch}':
+          publishers:
+            - ircbot_freenode
+            - discourse


theforeman.org/yaml/publishers/ircbot_freenode.yaml
```diff
@@ -29,3 +29,9 @@
             - name: '#theforeman-dev'
               notify-only: true
         matrix-notifier: only-configurations
+
+- publisher:
+    name: email_list
+    publishers:
+    - email:
+        recipients: ci@theforeman.org
```
ekohl, line 37:
```suggestion
        recipients: ci@community.theforeman.org
```

theforeman.org/yaml/publishers/ircbot_freenode.yaml
```diff
@@ -29,3 +29,9 @@
             - name: '#theforeman-dev'
               notify-only: true
         matrix-notifier: only-configurations
+
+- publisher:
+    name: email_list
```
ekohl, line 34: I'd prefer to use https://community.theforeman.org/c/infra/21
```suggestion
    name: discourse
```

theforeman.org/yaml/jobs/test_plugin.yaml
```diff
@@ -29,6 +29,7 @@
           block: true
     publishers:
       - ircbot_freenode
+      - email_list
```
ekohl, line 32: There's no need to modify all jobs. You can override the publishers in the bootdisk plugin config.

theforeman.org/yaml/publishers/ircbot_freenode.yaml
```diff
@@ -29,3 +29,9 @@
             - name: '#theforeman-dev'
               notify-only: true
         matrix-notifier: only-configurations
+
+- publisher:
```
ekohl, line 33: Please create a new file for this

ekohl (changes requested):

lzap:
Sure, so we have `ci_community.theforeman.org` (scrambled email) for that. Shall I simply create new publisher with this email address?

theforeman.org/yaml/jobs/test_plugin.yaml
lzap: Sure but woudn't it be more consistent to send all failures there? I think our community could benefit from this, unless it creates too much noise. These are executed weekly, we should be fine.

ekohl: I think it creates too much noise and would prefer to start with one. I don't want to end up with a board that has so many things there that nobody looks anymore.

ekohl:
> Sure, so we have `ci_community.theforeman.org` (scrambled email) for that. Shall I simply create new publisher with this email address?

Yes

theforeman.org/yaml/jobs/test_plugin.yaml
lzap: Let's enable that just for discovery and see how it works.

lzap:
Ok rebased, from the docs it looks like this is how am I supposed to redefine a template, but in the example there they only redefine a variable. I am not sure if this will work for publishers as well. Hopefully we have some tests.

theforeman.org/yaml/publishers/discourse.yaml
```diff
@@ -0,0 +1,5 @@
+- publisher:
+    name: email_list
```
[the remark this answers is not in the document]
lzap: Yeah sorry, renaming stuff and missed it. Fixed.

lzap:
Amended.

lzap:
I guess this was not needed. Thanks!

PR #111, opened 2021-09-28
fix syntax of provision playbook

evgeni:


ekohl (commented):
Is this why we have the `Retry geting list of nodes` block below or is that still needed to wait for the host to be provisioned (and poorly named)?

ekohl (approved):
Right, so that should be renamed (even if it was to fix the typo in getting :P) to something like `Wait until host is available`

PR #116, opened 2021-10-08
Upgrade version bump to 1.21

ezr-ondrej:
We no longer want to support upgrades from older version then 1.21 as the migrations won't be very reliable.
This bumps the tested version to the currently supported one.

This is mostly due to https://github.com/theforeman/foreman/pull/8790


ekohl:
I'm not sure there's a lot of value in that. From a developer perspective I see little difference and it may provide a bit more safety in allowing users to upgrade from older versions, even if we don't officially support that.

PR #118, opened 2021-10-11
force REQUESTS_CA_BUNDLE=/etc/pki/tls/cert.pem

evgeni:
we install jenkins-jobs via pip, which pulls in certifi, which contains
the old DST X3 root cert, which makes OpenSSL 1.0.x on EL7 freak out and
fail to verify the perfectly valid certs of ci.theforeman.org and
ci.centos.org


ehelms:
The commit message might could use a change from `foce` to `force` as well. More typing @evgeni !

PR #127, opened 2021-10-26
Stop building on Ubuntu 18.04 Bionic

ekohl:
This also switches the plugin builds from Bionic to Buster. I wasn't sure if we should switch from Ubuntu to Debian, but Buster is certainly supported on all versions for which we build now and is the oldest.


ehelms:
Since we want this for 3.1, and this is stabilization week I am going to merge this to start propagating changes.

evgeni (commented):
post-merge ACK

PR #128, opened 2021-10-27
Fix setting Jenkins job for Debian packaging PR jobs

ehelms:
When setting up a core build that pulls source from Jenkins, the
code will assume a default structure for the name of the Jenkins job
and for the job ID to use. Additionally, makes use of the build_vars.sh
concept to allow overriding of these defaults when they are different
and puts the power in the packaging repository to define values rather
than hard-coding into the CI.

My idea is then that for packages that differ from the pattern, e.g. foreman-proxy where its job name is `smart-proxy-develop-source-release` we can add to `build_vars.sh` the `job_name` variable. If that sounds cool, I'll open a PR to packaging to do that first.

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -186,11 +186,25 @@ def deb_build_dir(type, project, os = null) {
     } else if (type == 'plugin') {
         build_dir = "plugins/build-${project}"
     } else {
-        echo "Unsupported type specified: ${type}"
-        exit 1
+        error message: "Unsupported type specified: ${type}"
     }
 
     return build_dir
+}
+
+def read_build_vars(project) {
+    def data = [:]
+
+    dir("../${project}") {
+        if (fileExists('build_vars.sh')) {
+            readFile('build_vars.sh').split("\n").each { line ->
+              split_line = line.split('=', 2)
+              data[split_line[0]] = split_line[1]
+            }
+        }
+    }
+
+    return data
 }
 
 def setup_sources_deb(type, project, os, version, repoowner, pull_request) {
@@ -202,8 +216,7 @@ def setup_sources_deb(type, project, os, version, repoowner, pull_request) {
     } else if (type == 'plugin') {
         setup_sources_plugin(project, os, version, repoowner, pull_request)
     } else {
-        echo "Unsupported type specified: ${type}"
-        exit 1
+        error message: "Unsupported type specified: ${type}"
     }
 
 }
@@ -217,17 +230,15 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
 
     dir(build_dir) {
         if (version == 'nightly') {
+            def build_vars = read_build_vars(project)
             def url_base = 'https://ci.theforeman.org'
-            def job_name = nightly_jenkins_job
-            def job_id = nightly_jenkins_job_id
-            def job_url = "${url_base}/job/${job_name}/${job_id}"
+            def job_id = 'lastSuccessfulBuild'
+            def job_url = "${url_base}/job/${build_vars['source_location']}/${job_id}"
 
             sh """
               wget "${job_url}/artifact/*zip*/archive.zip"
               unzip archive.zip
               mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
-
-              # Set this in case we need it
             """
 
             last_commit = sh(returnStdout: true, script: "curl \"${job_url}/api/json\" | jq -r '.actions[].lastBuiltRevision.SHA1 | values'").trim()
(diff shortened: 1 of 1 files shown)


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -267,24 +279,21 @@ def setup_sources_dependency(project, os, version, repoowner, pull_request = fal
...
-            # Import variables from the project, allowing it to override behaviour
-            if [ -e ../${project}/build_vars.sh ]; then
-              . ../${project}/build_vars.sh
-            fi
+        def build_vars = read_build_vars("../${project}")
+        def build_type = build_vars.containsKey('BUILD_TYPE') ? build_vars['BUILD_TYPE'] : 'gem'
 
-            if [ "\$BUILD_TYPE" = "python" ]; then
+        sh """
+            if [ "${build_type}" = "python" ]; then
```
evgeni, line 286: given `build_type` is now a Groovy-level variable, should the `if/else` also become Groovy and the `sh` calls move *inside* the `if/else`? IMHO much saner to read, but also scope-creep somehow.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -267,24 +279,21 @@ def setup_sources_dependency(project, os, version, repoowner, pull_request = fal
...
 
     dir(build_dir) {
-        sh """
-            BUILD_TYPE=gem
-            # Import variables from the project, allowing it to override behaviour
-            if [ -e ../${project}/build_vars.sh ]; then
-              . ../${project}/build_vars.sh
-            fi
+        def build_vars = read_build_vars("../${project}")
+        def build_type = build_vars.containsKey('BUILD_TYPE') ? build_vars['BUILD_TYPE'] : 'gem'
```
evgeni, line 283: doesn't this work too?

```suggestion
        def build_type = build_vars.get('BUILD_TYPE', 'gem')
```

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -193,6 +193,19 @@ def deb_build_dir(type, project, os = null) {
     return build_dir
 }
 
+def read_build_vars(directory) {
+    def data = [:]
+
+    dir(directory) {
+        readFile('build_vars.sh').split("\n").each { line ->
```
evgeni, line 200: should this check for the presence of `build_vars.sh` first? I think it should.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -193,6 +193,19 @@ def deb_build_dir(type, project, os = null) {
...
+          split_line = line.split('=')
```
evgeni, line 201: we only want two entries in the resulting array
```suggestion
          split_line = line.split('=', 2)
```

evgeni (commented):
I've left a few inline comments, but overall a big :+1: on the aproach

While reviewing, two questions came to my mind:
1. the file is not (strictly speaking) a shell file anymore, should we drop (or replace) the `.sh` suffix? this would require changes to stable branches *OR* some backward compat code (checking for both file names for a while), but I think it'd make things clearer
2. Right now, all calls read like `read_build_vars("../${project}")` as that's the directory where the file is to be found. If we make this `read_build_vars(project)` instead, and then do the `"../${project}/build_vars.sh"` logic *inside* the method, we could easily have the next step be replacing per-project files with a big `debian_manifest.yaml` in the repository root and having the vars for each project being set there

Nothing of the :point_up: should be blocking or even the exact path we should go after this PR, but some food for thought ;)

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -217,17 +230,16 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
...
     dir(build_dir) {
         if (version == 'nightly') {
+            def build_vars = read_build_vars("../${project}")
             def url_base = 'https://ci.theforeman.org'
-            def job_name = nightly_jenkins_job
-            def job_id = nightly_jenkins_job_id
-            def job_url = "${url_base}/job/${job_name}/${job_id}"
+            def job_name = "${project}-develop-source-release"
+            def job_id = 'lastSuccessfulBuild'
+            def job_url = "${url_base}/job/${build_vars['source_location']}/${job_id}"
```
evgeni, line 237: And because I forgot to mention this in the first round: should this be `nightly_source_location`, or even `nightly_source_job` just to explicitly point out it's used for nightly only?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -217,17 +230,16 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
 
     dir(build_dir) {
         if (version == 'nightly') {
+            def build_vars = read_build_vars("../${project}")
             def url_base = 'https://ci.theforeman.org'
-            def job_name = nightly_jenkins_job
-            def job_id = nightly_jenkins_job_id
-            def job_url = "${url_base}/job/${job_name}/${job_id}"
+            def job_name = "${project}-develop-source-release"
```
evgeni, line 235: this is now unused?

ehelms:
> I've left a few inline comments, but overall a big +1 on the aproach
> 
> While reviewing, two questions came to my mind:
> 
> 1. the file is not (strictly speaking) a shell file anymore, should we drop (or replace) the `.sh` suffix? this would require changes to stable branches _OR_ some backward compat code (checking for both file names for a while), but I think it'd make things clearer

If I do this, it feels like I might be moving towards having a top level file anyway as I'll need compat code.

> 2. Right now, all calls read like `read_build_vars("../${project}")` as that's the directory where the file is to be found. If we make this `read_build_vars(project)` instead, and then do the `"../${project}/build_vars.sh"` logic _inside_ the method, we could easily have the next step be replacing per-project files with a big `debian_manifest.yaml` in the repository root and having the vars for each project being set there

I thought about this and flipped a coin where I ended up. I did not think of the angle of that making the interface more long term compatible.

> 
> Nothing of the point_up should be blocking or even the exact path we should go after this PR, but some food for thought ;)

theforeman.org/pipelines/lib/packaging.groovy
ehelms: `source_location` is what we use with RPM variables so I wanted to keep consistency

ehelms:
Pushed changes based on the inline comments and moved the method signature of read_build_vars to just take a project.

theforeman.org/pipelines/lib/packaging.groovy
evgeni: ok

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -318,30 +330,29 @@ def setup_sources_plugin(project, os, version, repoowner, pull_request = false)
...
                     tar -x -C "${project}" -f "${project}_${package_version}.orig.tar.gz"
-                else
-                    echo "Unsupported build type: \${BUILD_TYPE}"
-                    exit 1
-                fi
-            """
+                """
+            } else {
+                echo "Unsupported build type: ${build_type}"
+                sh "exit 1"
```
evgeni, line 354:
```suggestion
                error message: "Unsupported build type: ${build_type}"
```

see https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#error-error-signal

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -267,27 +280,26 @@ def setup_sources_dependency(project, os, version, repoowner, pull_request = fal
...
               ../../gem2deb ${project}-${package_version}.gem --debian-subdir ../${project} --only-source-dir
-            else
-              echo "Unsupported build type: \${BUILD_TYPE}"
-              exit 1
-            fi
-        """
+            """
+        } else {
+            echo "Unsupported build type: ${build_type}"
+            sh 'exit 1'
```
evgeni, line 301: same as below (https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#error-error-signal)

evgeni (approved):

PR #129, opened 2021-11-02
Unpin bundler

tbrisker:
We pinned bundler to < 2.0 in 71d247e1933 but forgot to unpin it now
that we don't need to support Ruby 2.2 or older.


evgeni (approved):
I am not sure all of these file are even still used, but this looks clean to me, and easy to revert if anything breaks.

ekohl (approved):
My only concern with this is foreman-maintain which still uses Ruby 2.0. However, I'm not sure we even test it in Jenkins.

evgeni:
> My only concern with this is foreman-maintain which still uses Ruby 2.0. However, I'm not sure we even test it in Jenkins.

FM only uses GHA for tests, we're safe.

PR #130, opened 2021-11-02
Separate scratchbuild from timestamp with a plus

ehelms:


ehelms:
Also would align with how the current timestamp using the old jobs is generated.

PR #132, opened 2021-11-03
inject nightly changelog entry when building nightly

evgeni:


evgeni:
@ehelms @ekohl this should fix PR-merges and nightly-builds of `deb/develop` not getting the proper version treatment.

ekohl (approved):

PR #133, opened 2021-11-03
Use the Copy Artifacts Plugin in packaging

ekohl:
This uses the [Copy Artifacts Plugin](https://plugins.jenkins.io/copyartifact/) (already installed) to download artifacts from another job.

It doesn't pass the selector since it defaults to lastSuccessful and it doesn't list lastSuccessfulBuild as a valid option.

However, that does make me wonder if that's correct. Shouldn't it be the upstream job in some cases or is the job already completed by the time this job runs?

Sadly a lot of the logic has to remain to get the last built revision.

Currently untested.

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -231,17 +231,9 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
     dir(build_dir) {
         if (version == 'nightly') {
             def build_vars = read_build_vars(project)
-            def url_base = 'https://ci.theforeman.org'
-            def job_id = 'lastSuccessfulBuild'
-            def job_url = "${url_base}/job/${build_vars['source_location']}/${job_id}"
-
-            sh """
-              wget "${job_url}/artifact/*zip*/archive.zip"
-              unzip archive.zip
-              mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
-            """
-
-            last_commit = sh(returnStdout: true, script: "curl \"${job_url}/api/json\" | jq -r '.actions[].lastBuiltRevision.SHA1 | values'").trim()
+            copyArtifacts(projectName: build_vars['source_location'], excludes: 'package-lock.json')
+            sh "mv *.tar.bz2 ${project}_${package_version}.orig.tar.bz2"
+            last_commit = readFile('commit').trim()
         } else {
             sh """
               # Download sources


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -235,11 +235,8 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
...
 
-            sh """
-              wget "${job_url}/artifact/*zip*/archive.zip"
-              unzip archive.zip
-              mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
-            """
+            copyArtifacts(projectName: build_vars['source_location'], filter: '*.tar.bz2')
+            sh "mv *.tar.bz2 ${project}_${package_version}.orig.tar.bz2"
 
             last_commit = sh(returnStdout: true, script: "curl \"${job_url}/api/json\" | jq -r '.actions[].lastBuiltRevision.SHA1 | values'").trim()
```
evgeni, line 241:
```suggestion
            copyArtifacts(projectName: build_vars['source_location'])
            sh "mv *.tar.bz2 ${project}_${package_version}.orig.tar.bz2"

            last_commit = readFile('commit').trim()
```

evgeni (approved):
LGTM!

Feel free to merge and kick off a run to see if it works as we expect it to?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -231,17 +231,9 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
...
-            def job_url = "${url_base}/job/${build_vars['source_location']}/${job_id}"
-
-            sh """
-              wget "${job_url}/artifact/*zip*/archive.zip"
-              unzip archive.zip
-              mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
-            """
-
-            last_commit = sh(returnStdout: true, script: "curl \"${job_url}/api/json\" | jq -r '.actions[].lastBuiltRevision.SHA1 | values'").trim()
+            copyArtifacts(projectName: build_vars['source_location'], excludes: 'package-lock.json')
```
ekohl, line 234: I didn't know if I could use `pattern` with a comma to separate. It's supposed to be the ant file pattern, but that wasn't very clear to me. This at least filters out the 1.5 MB package-lock.json file from Foreman's artifacts. Bundler's lock file is really tiny so that doesn't matter much.

PR #137, opened 2021-11-05
always take the last lastBuiltRevision

evgeni:
sometimes, a source-release job is triggered by multiple merges, which
results in multiple BuildData objects being present in the JSON output

the old jq query printed *all* SHA1 of these, resulting in bad Debian
changelogs being generated

lets take only the last entry and print its SHA1

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -241,7 +241,7 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
               mv archive/pkg/*bz2 ${project}_${package_version}.orig.tar.bz2
             """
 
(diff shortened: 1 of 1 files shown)


evgeni:
if you're curious (I know you are!), one such job was: https://ci.theforeman.org/job/foreman-develop-source-release/1404/

ekohl:
I wondered if there was a native way of doing this. In https://github.com/theforeman/jenkins-jobs/pull/133 I wondered about this. Perhaps we can create a dedicated archive file for the commit?

evgeni:
Funfact, we actually do!

https://ci.theforeman.org/job/foreman-develop-source-release/1404/artifact/commit

evgeni:
So I guess this is the way? https://github.com/theforeman/jenkins-jobs/pull/133/files#r743590778

PR #141, opened 2021-11-09
Adding Pipeline Definitions for Foreman 3.1

Odilhao:


theforeman.org/pipelines/vars/foreman/3.1.groovy
```diff
@@ -0,0 +1,41 @@
...
+def foreman_el_releases = [
+    'el7',
+    'el8'
+]
+def foreman_debian_releases = ['buster', 'bionic', 'focal']
+
+def pipelines_deb = [
+    'install': [
+        'debian10',
+        'ubuntu1804',
```
upadhyeammit, line 17: we are dropping ubuntu1804, so I think we dont require 'ubuntu1804'? @ekohl am I correct?

theforeman.org/pipelines/vars/foreman/3.1.groovy
```diff
@@ -0,0 +1,41 @@
...
+        'ubuntu2004'
+    ],
+    'upgrade': [
+        'debian10',
+        'ubuntu1804'
```
upadhyeammit, line 22: ditto @ekohl am I correct?

centos.org/jobs/foreman-pipelines.yml
```diff
@@ -16,6 +16,7 @@
       - '2.4'
       - '2.5'
       - '3.0'
+      - '3.1'
```
ekohl, line 19: This will include Ubuntu 18.04. You should modify the nightly config below. I'm not sure about the naming

ekohl (changes requested):

centos.org/jobs/foreman-pipelines.yml
Odilhao: I already removed the Ubuntu 18.04 `theforeman.org/pipelines/vars/foreman/3.1.groovy` , isn't the VAR supposed to bring only the SO that is declared inside the 3.1.groovy file?

ekohl: Technically you can add it to `foreman-nightly` and rename it to `foreman-31-and-newer`. The name isn't really used anyway.

Odilhao: I already did on  d2fe358841ae234a43a670b080cc2e069f137f61 , I can change it before the rebase/squash

ekohl (approved):
We can clean up later. Looks like we still have 2.3 and 2.4 but those are EOL. Good follow up.

evgeni:
> We can clean up later. Looks like we still have 2.3 and 2.4 but those are EOL. Good follow up.

Yeah, we kept 2.3 around due to Katello 3.18, but this is dead now, so it can goooooooooooooooo.

Odilhao:
I don't have merge permission :sweat_smile:

PR #143, opened 2021-11-12
fetch only the real debug files

evgeni:


evgeni:
(draft, as I still need to test this)

centos.org/ansible/fetch_debug_files.yml
```diff
@@ -3,7 +3,7 @@
   tasks:
     - name: 'Find any files from debug'
       find:
-        paths: '/tmp'
+        paths: "/tmp/debug-{{ pipeline_type | default('foreman') }}-{{ pipeline_version | default('nightly') }}-{{ pipeline_os | default('el7') }}"
```
ekohl, line 6: I'm not sure if defaults make sense here. Would `*` be a better default to then just capture it all? Or fail if they're undefined?

ekohl (approved):
Don't consider my comments as blocking.

centos.org/ansible/fetch_debug_files.yml
evgeni: `paths` needs to be a real path, not a wildcard, so `*` wouldn't work, sadly.

I copied the defaults from https://github.com/theforeman/forklift/blob/60bbe0b1ff74e8375909866f823926e8d93e8b51/playbooks/collect_debug.yml#L6

there are set there as that playbook can be used outside the pipeline context (and they match the defaults in the bats role: https://github.com/theforeman/forklift/blob/60bbe0b1ff74e8375909866f823926e8d93e8b51/roles/bats/defaults/main.yaml#L9)

but you have a point, this playbook isn't used outside the pipeline context, so the `pipeline_*` vars will always be defined.

PR #145, opened 2021-11-15
search for cico plugin in both Python 2.7 and Python 3.6 installs

evgeni:
new el8 cico pods use python 3.6


ekohl (approved):
I assume we're not really rid of EL7 just yet and can't change that to Python 3.6 either.

evgeni:
> I assume we're not really rid of EL7 just yet and can't change that to Python 3.6 either.

Correct, the old jobs still run on EL7 with Python 2, once we completely move to the new OpenShift setup, we can remove those.

PR #146, opened 2021-11-15
CentOS OCP migration

evgeni:


centos.org/ansible/provision.yml
```diff
@@ -8,40 +8,40 @@
       register: cico_get_data
       retries: 2
       delay: 310
-      until: cico_get_data.results is defined
+      until: cico_get_data.ansible_module_results is defined
```
evgeni, line 11: this will need another change once https://github.com/CentOS/python-cicoclient/commit/d7cda54e269e7cfd5eee156d1f6b0463d3b4d861 is released, but then it should be stable

centos.org/ansible/templates/ssh_config.j2
```diff
@@ -1,10 +1,11 @@
...
+{% for host, value in cico_data.ansible_module_results.hosts.items() %}
 Host duffy_box
   HostName {{ value.hostname  }}
   User root
   Port 22
   StrictHostKeyChecking no
   PasswordAuthentication no
-  GlobalKnownHostsFile=/dev/null
+  GlobalKnownHostsFile /dev/null
+  IdentityFile /duffy-ssh-key/ssh-privatekey
```
evgeni, line 9: This is set in `/etc/ssh/ssh_config` on in the container, but as we use `ssh -F`, the global file isn't read :(

theforeman.org/pipelines/lib/foreman_infra.groovy
```diff
@@ -10,7 +10,7 @@ void git_clone_jenkins_jobs(args = [:]) {
     target_dir = args.target_dir ?: ''
 
     dir(target_dir) {
-        git url: 'https://github.com/theforeman/jenkins-jobs', poll: false
+        git url: 'https://github.com/evgeni/jenkins-jobs', branch: 'centos-ocp', poll: false
```
evgeni, line 13: this can go, I just needed it here so it pulls the ansible playbooks from my fork

centos.org/ansible/templates/ssh_config.j2
ekohl: Thinking out loud: can we concat `/etc/ssh/ssh_config` and this template to avoid hardcoding the path?

centos.org/ansible/provision.yml
```diff
@@ -41,7 +41,7 @@
       copy:
         content: |
           {% for host, value in cico_data.ansible_module_results.hosts.items() %}
-          {{ value.hostname }} ansible_fqdn={{ value.hostname }} ansible_user=root ansible_ssh_private_key_file=~/.ssh/id_rsa ansible_host={{ value.ip_address }}
+          {{ value.hostname }} ansible_fqdn={{ value.hostname }} ansible_user=root ansible_ssh_private_key_file=/duffy-ssh-key/ssh-privatekey ansible_host={{ value.ip_address }}
```
ekohl, line 44: If we don't specify the private key file, does it respect the SSH config that's already in place?

evgeni: Yeah, it probably does.

centos.org/ansible/templates/ssh_config.j2
evgeni: Not trivially, but I can try a few things.

evgeni: The global config is btw here: https://github.com/centosci/images/blob/master/cico-workspace/ssh_config

evgeni:
blocked by https://pagure.io/centos-infra/issue/728

centos.org/ansible/templates/ssh_config.j2
evgeni: I think I've managed to make it nicely ;)

ehelms (approved):

evgeni:
needed only one follow up: https://github.com/theforeman/jenkins-jobs/pull/212/

WOW ;)

PR #154, opened 2021-11-26
nightly: release bullseye packages

evgeni:


evgeni:
those packages are usable, if the user has *buster* `puppet-agent` and `puppetserver` available

ehelms (approved):

PR #155, opened 2021-11-26
nightly: test bullseye

evgeni:
this will currently fail, as there are no bullseye packages for `puppetserver` and our pipelines don't apply silly hacks (usually)


evgeni:
workaround implemented in https://github.com/theforeman/forklift/pull/1448

ehelms (approved):

PR #156, opened 2021-12-01
Set JEST_TIMEOUT for katello react-ui

jturel:
This moves the 300 second timeout to CI only and out of dev so that we can keep a better eye on long-running tests


jturel:
@evgeni can you take a look when you get a sec? :)

PR #158, opened 2021-12-03
drop Foreman 2.4 and Katello 4.0

evgeni:
contains #157 for easier rebasing


evgeni:
cc @upadhyeammit - 2.4 is "dead" by now, right? we don't plan any more releases there?
cc @jjeffers - 4.0 is "dead" by now, right? we don't plan any more releases there?

jjeffers:
Yeah 4.0 is dead, AFAIK

upadhyeammit:
> cc @upadhyeammit - 2.4 is "dead" by now, right? we don't plan any more releases there? cc @jjeffers - 4.0 is "dead" by now, right? we don't plan any more releases there?

Considering we can still do the release of [2.4.2](https://da.gd/tfm-242) **if required** before or by 3.1 GA. However if I see the fixes in 2.4.2 as of now then those are not high priority apart from [#14409](https://projects.theforeman.org/issues/14409) which is not reproducible all time. As per this we are not doing 2.4.2. But do we need 2.4 to test upgrades to 3.0.z? If I remember then we try the upgrades from N-2 in release pipelines or that has changed?

evgeni:
This is only for building/releasing, n-2 for upgrades is not affected by these changes.

upadhyeammit:
> This is only for building/releasing, n-2 for upgrades is not affected by these changes.

Then go ahead!

ehelms (approved):

PR #166, opened 2022-01-05
Pass environmental variables into FDI pipeline

lzap:
When vagrant is started to build FDI, it previously accepted three environmental variables. This was never implemented in this new pipeline so FDI was always built from master as nightly (default values):

https://github.com/theforeman/foreman-discovery-image/blob/master/aux/vagrant-build/Vagrantfile

Since passing env variables from ssh_duffy is clunky, I am making a change to accept regular arguments:

https://github.com/theforeman/foreman-discovery-image/blob/cos8-novirt/aux/vagrant-build/Vagrantfile

This PR changes passing of options to match the change, however, I am not sure if these variables are correctly passed into this job. Where do I set this?

Thanks


lzap:
I edited what I think are the inputs, not sure if these gets passed/rendered into `${env.xyz}`

centos.org/pipelines/foreman-discovery-image-build.groovy
```diff
@@ -32,7 +33,7 @@ pipeline {
             steps {
                 script {
                     duffy_ssh("git clone https://github.com/theforeman/foreman-discovery-image/ --branch ${env.branch}", 'duffy_box', './')
```
ekohl, line 35: Shouldn't this also respect the owner?

centos.org/pipelines/foreman-discovery-image-build.groovy
```diff
@@ -32,7 +33,7 @@ pipeline {
...
-                    duffy_ssh("cd foreman-discovery-image/aux/vagrant-build/ && vagrant up fdi-builder", 'duffy_box', './')
+                    duffy_ssh("cd foreman-discovery-image/aux/vagrant-build/ && vagrant --repo-owner=${env.repo_owner} --branch=${env.branch} --proxy-repo=${env.proxy_repository} up fdi-builder", 'duffy_box', './')
```
ekohl, line 36: How does this work? I'm not aware of vagrant accepting custom parameters.

lzap: See here: https://github.com/theforeman/foreman-discovery-image/blob/cos8-novirt/aux/vagrant-build/Vagrantfile

centos.org/pipelines/foreman-discovery-image-build.groovy
lzap: Good catch, fixed.

lzap:
I merged the change into FDI repo, this should be ready to merge: https://github.com/theforeman/foreman-discovery-image/commit/034e6fa798c13463c2414bae7eca6d665ad12e4b

centos.org/pipelines/foreman-discovery-image-build.groovy
lzap: Ok I reverted it back to ENV variables: https://github.com/theforeman/foreman-discovery-image/commit/1b46f13bea3b4d80e2c3929aabbf2e64d9aa0e7b

lzap:
Rebased, this is ready for re-review.

centos.org/pipelines/foreman-discovery-image-build.groovy
```diff
@@ -45,6 +46,7 @@ pipeline {
         success {
             archiveArtifacts artifacts: 'result/fdi*tar', allowEmptyArchive: true
             archiveArtifacts artifacts: 'result/fdi*iso', allowEmptyArchive: true
+            archiveArtifacts artifacts: 'result/*log', allowEmptyArchive: true
```
lzap, line 49: I would like to see log files in the artifacts as well.

ekohl (approved):

PR #167, opened 2022-01-13
throw away output of db:drop where we also ignore the exit code

evgeni:


lzap:
Reviewers: You can read about my complain and motivation for this in here: https://github.com/theforeman/jenkins-jobs/pull/168

ekohl (approved):

PR #171, opened 2022-01-18
Add back 3.18 pr pipeline

jlsherrill:


jlsherrill:
i'm adding this back because both Atix and us would like to contribute fixes to the 3.18 branch, but the tests no longer run there.  We have no plans for another 3.18 release though

ekohl (approved):
I don't mind this. If all this costs is a slightly longer mapping, it's totally worth it.

evgeni:
Just realized, this means we need to pick the "pin rdoc to < 6.4" change to foreman 2.3-stable for that to actually work.

Absolutely doable, just nothing I remembered when we talked on IRC about this.

evgeni:
I wasn't sure if Bundler doesn't complain if a dep is defined twice.
pip does.

jlsherrill:
as long as the requires don't conflict, it should be fine.  We'll see!

evgeni:
Seems to have passed, yay

PR #174, opened 2022-02-07
Adding Pipeline Definitions for Foreman 3.2

zjhuntin:


centos.org/jobs/foreman-pipelines.yml
```diff
@@ -41,6 +41,26 @@
...
       - plugins
 
+- project:
+    name: foreman-release-test-32
+    jobs:
+      - 'foreman-pipeline-{type}-{version}-{distro}-{action}'
+    empty: ''
+    distro:
+      - centos7
+      - centos8
```
pcreech, line 51: I think we want to remove centos8 at this point?

centos.org/jobs/foreman-pipelines.yml
```diff
@@ -41,6 +41,26 @@
...
+      - centos8-stream
+      - debian10
```
pcreech, line 53: nightly has debian11.as well.  And the 3.2.groovy attached to this PR also has debian11.  I think we need to also add debian11 here.

PR #176, opened 2022-02-09
Add release view templates

### LINKED ISSUE #169: templated release test jobs
We should also change the `test_$release` and `test_proxy_$release` jobs to pipelines fitting the naming scheme. So `foreman-1.24-stable-source` and `smart-proxy-1.24-stable-source`. Again, these should be driven by `yaml/jobs/pipeline/foreman-release.yaml`.

ekohl:
This adds templated views for releases which makes it easier to have the correct version tabs for a quick overview.

Currently untested. I've followed https://docs.openstack.org/infra/jenkins-job-builder/definition.html#views but I don't know if this will start to purge unmanaged views.

Fixes #169

### DIFF
--- a/theforeman.org/yaml/views/release.yml
+++ b/theforeman.org/yaml/views/release.yml
@@ -1,0 +1,22 @@
+- view-template:
+    name: 'Foreman {major}.{minor}'
+    description: 'Foreman {major}.{minor} jobs'
+    view-type: list
+    regex: '.*{major}[_\.]{minor}.*'
+
+- project:
+    name: 'foreman-releases'
+    views:
+        - 'Foreman {major}.{minor}'
+    major:
+        - '3'
+    minor:
+        - '0'
+        - '1'
+        - '2'
+
+- view:
+    name: 'Foreman Nightly'
+    description: 'Foreman nightly jobs'
+    view-type: list
+    regex: '.+-((nightly|plugins)-(rpm|deb)-pipeline|(develop|master|source|package)-(test|release))'


theforeman.org/yaml/views/release.yml
```diff
@@ -0,0 +1,21 @@
...
+    views:
+        - '{name}-template-{major}.{minor}'
+    major:
+        - '3'
+    minor:
+        - '0'
+        - '1'
+
+- view:
+  name: 'Foreman Nightly'
```
ehelms, line 18: Looks like this line and each below need to be indented

theforeman.org/yaml/views/release.yml
```diff
@@ -0,0 +1,21 @@
+- view-template:
+    name: 'Foreman {major}.{minor}'
+    description: 'Foreman {major}.{minor} jobs'
+    view-type: list
+    regex: '.*{major}[_\.]{minor}.*'
+
+- project:
+    name: 'foreman-releases'
+    views:
+        - '{name}-template-{major}.{minor}'
```
ehelms, line 10: I think this needs to match the `name:` defined on https://github.com/theforeman/jenkins-jobs/pull/176/files#diff-b6bd4b212bdd5eae7e1f14909bf3b48d97db0a2b5d14afd613c56b60ddc54528R2 in order for it to link them up properly

ekohl:
At least the tests are green now. When we deploy this, I don't know whether it'll remove all non-managed views or if we don't purge those.

ehelms:
We run `update --delete-old -r .` which to my understanding *only* purges jobs and views that were previously managed by JJB. Given none of our views are currently, I do not think this will delete any of the existing views.

> The update command includes a delete-old option to remove obsolete jobs:

jenkins-jobs update --delete-old /path/to/defs
Obsolete jobs are jobs once managed by JJB (as distinguished by a special comment that JJB appends to their description), that were not generated in this JJB run.

ehelms (approved):

ekohl:
Doesn't look like it's starting to manage views:
https://ci.theforeman.org/job/jenkins-jobs-update/165/console
However, it did fail on one job. Looking at the Jenkins logs:
> WARNING hudson.triggers.Trigger#checkTriggers: The job katello-pr-test has a syntactically incorrect config and is missing the cron spec for a trigger

So it probably doesn't proceed with views until that's fixed.

ekohl:
Ok, so I had to:
* Edit katello-pr-test and save it in the UI
* Add permissions for the JJB user to manage views

Now we have views:
https://ci.theforeman.org/view/Nightly/ vs https://ci.theforeman.org/view/Foreman%20Nightly/

The former does have a list of unstable jobs, which can be quite useful. And the description may not be that useful and only takes up space.

ehelms:
> The former does have a list of unstable jobs, which can be quite useful

Is this something that can be denoted in JJB to include?

Now that this is working - is our follow up to put the rest of the views into code?

ekohl:
> Is this something that can be denoted in JJB to include?

I haven't checked yet

> Now that this is working - is our follow up to put the rest of the views into code?

Yes, I do think that would be a good thing. It would remove yet another piece of manually managed Jenkins.

PR #163, opened 2022-01-03
stop testing nightly on centos 8

evgeni:
centos 8 is EOL and the repos will go away soon too

PR #178, opened 2022-02-11
Add Katello 4.4 pipeline

ianballou:
Adds the Katello 4.4 pipeline and removes Katello 4.1 since it's the oldest.

### DIFF
--- a/centos.org/jobs/katello-pipelines.yml
+++ b/centos.org/jobs/katello-pipelines.yml
@@ -41,6 +41,21 @@
     distro:
       - centos7
       - centos8
+    action:
+      - install
+      - upgrade
+    type:
+      - katello
+
+- project:
+    name: katello-release-4.4
+    jobs:
+      - 'foreman-pipeline-{type}-{version}-{distro}-{action}'
+    empty: ''
+    version:
+      - '4.4'
+    distro:
+      - centos7
       - centos8-stream
     action:
       - install
--- a/theforeman.org/pipelines/vars/katello/4.4.groovy
+++ b/theforeman.org/pipelines/vars/katello/4.4.groovy
@@ -1,0 +1,16 @@
+def foreman_version = '3.2'
+def katello_version = '4.4'
+def foreman_el_releases = [
+    'el7',
+    'el8'
+]
+def pipelines = [
+    'install': [
+        'centos7',
+        'centos8-stream',
+    ],
+    'upgrade': [
+        'centos7',
+        'centos8-stream',
+    ]
+]


theforeman.org/pipelines/vars/katello/4.4.groovy
```diff
@@ -0,0 +1,17 @@
...
+]
+def pipelines = [
+    'install': [
+        'centos7',
+        'centos8',
+        'centos8-stream',
+    ],
+    'upgrade': [
+        'centos7',
+        'centos8',
```
ianballou, line 15: Should 'centos8-stream' be in here too now?  I'm unsure because I think the upgrade pipeline goes back two versions, which would bring us to 4.2...

ianballou:
I may have opened this up a bit too early :)

centos.org/jobs/katello-pipelines.yml
```diff
@@ -16,28 +16,29 @@
...
+    name: katello-release-4.3
     jobs:
       - 'foreman-pipeline-{type}-{version}-{distro}-{action}'
     empty: ''
     version:
-      - '4.2'
+      - '4.3'
     distro:
       - centos7
       - centos8
```
ekohl, line 27: You should probably even drop this. The CentOS Linux 8 images no longer work so you'll run into that when you want to release the next 4.3.z. Same applies to 4.2 and 4.4.

```suggestion
```

ekohl (changes requested):
When we branch, we always look at nightly. In this case you can see nightly doesn't have plain `centos8` anymore.

See eb2960cb0f668a6364c051385cd15eae89065b7f / https://github.com/theforeman/jenkins-jobs/pull/163 as well.

ianballou:
Okay, I dropped `centos8` as a distro in the pipelines.

ianballou:
I edited so that the oldest release isn't dropped so soon.

centos.org/jobs/katello-pipelines.yml
```diff
@@ -8,7 +8,6 @@
       - '4.1'
     distro:
       - centos7
-      - centos8
```
ekohl, line 11: I'm not sure we should drop this. I'm leaning to first a commit that moves everything from centos8 to centos8 stream, then a commit that adds 4.4.

theforeman.org/pipelines/vars/katello/4.4.groovy
```diff
@@ -0,0 +1,15 @@
...
+    'el7',
+    'el8'
+]
+def pipelines = [
+    'install': [
+        'centos7',
+        'centos8-stream',
+    ],
+    'upgrade': [
+        'centos7',
```
ekohl, line 13: We can also upgrade on Stream, right? We can install 4.2, then move to 4.3 and then 4.4.

ianballou: I left it out because I didn't see `centos8-stream` in the 4.2 install pipeline.  If that doesn't matter then I can add in `centos8-stream`.

ekohl (approved):

PR #179, opened 2022-02-11
Add Katello 4.4 + Foreman 3.2 test mapping

ianballou:
Katello 4.4 is branching today.


ekohl (commented):
For future reference, you can submit a PR to just add 4.4 / 3.2 and that can be merged even before branching. An additional mapping doesn't hurt.

With, is it also certain there won't be any 4.1.z release anymore?

ianballou:
I'm double checking about future 4.1 releases. I'm pretty certain we won't have any more though.

ianballou:
Okay, yeah, no more 4.1 releases are planned.

ianballou:
> For future reference, you can submit a PR to just add 4.4 / 3.2 and that can be merged even before branching. An additional mapping doesn't hurt.

I can add something in tool_belt to suggest doing this earlier.

ekohl:
In Foreman we typically have a policy where 2 releases are supported and you can have 1 in RC stages. So now 3.0 and 3.1 are supported while 3.2 will soon be RC1. Once 3.2 goes GA, 3.0 is EOL. There can also be unexpected releases, such as with a security update. That's why Foreman only has instructions to add, not to remove. Removal happens after the GA. I'd recommend doing the same.

Right now we don't have it in our procedures, but you can see there are some checks for GA vs RC vs .z releases:

https://github.com/theforeman/tool_belt/blob/master/procedures/foreman/release.md.erb#L112-L123

That would probably be the place to capture it.

ianballou:
@ekohl that sounds fine, so I'll just add in the new 4.4 definitions for now and then the older ones can be removed later when 4.4 is GA'd.  

Our branching docs say to remove the oldest one even before the GA release, so I'll fix that.

ekohl:
Excellent. I like that we're aligning more and more processes.

ekohl (approved):

PR #192, opened 2022-03-11
allow the GHA bot to be built automatically

evgeni:


evgeni:
the name is in line with how it created the staging repo, so I think it should be correct: https://stagingdeb.theforeman.org/pool/plugins/github-actions%5bbot%5d/

ehelms (approved):

PR #193, opened 2022-03-11
use both notations of "github-actions"

evgeni:
just to be sure, the former didn't seem to work, but given that name is
used everywhere, let's keep them so it won't break in future


evgeni:
proof it triggered: https://github.com/theforeman/foreman-packaging/pull/7699

ehelms (approved):

PR #194, opened 2022-03-14
bump timeouts of pipelines with plugins

evgeni:
the upgrade pipes have been hitting the limits recently, and we should
really not throw away the results in such a case


ekohl (approved):
I wonder why they need so much more time now. Has our install gotten slower?

evgeni:
> I wonder why they need so much more time now. Has our install gotten slower?

I think we added some intermediate steps to the tests that re-run installer and things.

PR #195, opened 2022-03-18
Drop Pulpcore 3.7, 3.9 & 3.11 pipelines

ekohl:
Katello 4.0 uses Pulpcore 3.14. While Katello 4.0 is unsupported, it can still be useful to keep around for blocking upgrades. However, all 3.x releases used 3.7 and that can now be considered unsupported. 3.9 and 3.11 were never used by Katello.


evgeni:
I wonder if there are still fixes released for 3.7 as that was the "migrate from pulp2" release. @ianballou might know.

Additionally, we also have galaxy ng as a consumer of those packages. Not sure which branches they still need in support. @pcreech @odilhao?

ianballou:
I'm really hoping it's not going to happen, but since Pulpcore 3.7 is related to the migration, it's possible that we'd need to release updates packages in case some emergency comes up. If it's not difficult to recreate a pipeline then I suppose we can drop it.

evgeni:
It's just the one line in the config, so not hard at all :)

evgeni:
There was an outstanding update for python-pulp_2to3_migration (0.11.5 to 0.11.6) in 3.7, which I pushed out. Merging now.

evgeni (approved):

PR #202, opened 2022-04-11
Drop EL7 from pipelines and jobs

upadhyeammit:


ehelms:
Handled here - https://github.com/theforeman/jenkins-jobs/pull/228

PR #205, opened 2022-04-14
add centos8-stream to katello 4.2 pipelines

evgeni:


evgeni:
@zjhuntin this should fix the failure in https://ci.theforeman.org/job/katello-4.2-rpm-pipeline/16/

centos.org/jobs/katello-pipelines.yml
```diff
@@ -25,6 +25,7 @@
     distro:
       - centos7
       - centos8
```
ekohl, line 27: Will this still work? Shouldn't this be dropped or do we have hacks to provide it with a repository (like from vault)?

ekohl (approved):

PR #206, opened 2022-04-14
clean old katello pipeline definitions

evgeni:


ekohl (approved):
What's the difference between 4.3, 4.4 and nightly now? Couldn't they be merged in a single entry? I'm also fine with keeping that for a future iteration.

evgeni:
Probably none, anymore. But yeah, later ;)

PR #207, opened 2022-04-19
add pulpcore 3.18 pipelines for el8 and el9

evgeni:


theforeman.org/pipelines/vars/pulpcore/version.groovy
```diff
@@ -7,3 +7,13 @@ def pipelines = [
...
+
+if (pulpcore_version == '3.18') {
+    pulpcore_distros = ['el8', 'el9']
+    pipelines = [
+        'pulpcore': [
+            'centos8-stream',
+            'centos9-stream'
+        ]
+    ]
+}
```
evgeni, line 19: this isn't pretty, but was the easiest way to add it without having a copy of the very same file for every pulpcore version BUT 3.18…

PR #220, opened 2022-06-08
Generate package-lock and run npm ci separately for Katello

ehelms:
This will hopefully provide some further granularity to understand
time spent generating dependency tree and downloading the NPM modules
in Katello PR and source jobs.


ehelms:
I threw the same idea into https://github.com/Katello/katello/pull/10154 as an idea to test.

evgeni (approved):
This matches what we talked about, so let's try it!

PR #221, opened 2022-06-09
purge useless npm deps before installing

evgeni:
requires https://github.com/theforeman/foreman/pull/9255 and is completely untested :)

### DIFF
--- a/theforeman.org/pipelines/test/testKatello.groovy
+++ b/theforeman.org/pipelines/test/testKatello.groovy
@@ -106,6 +106,16 @@ pipeline {
                     stage('assets-precompile') {
                         steps {
                             dir('foreman') {
+                                sh "python script/filter-package-json.py"
+
+                                withRVM(["bundle exec ruby script/plugin_webpack_directories.rb > plugin_webpack.json"], ruby)
+                                script {
+                                    def plugin_webpack = readJSON file: 'plugin_webpack.json'
+                                    plugin_webpack['plugins'].each { plugin, config ->
+                                        sh "python script/filter-package-json.py --package-json ${config['root']}/package.json"
+                                    }
+                                }
+
                                 sh "cp db/schema.rb.nulldb db/schema.rb"
                                 withRVM(["bundle exec npm install --package-lock-only --no-audit"], ruby)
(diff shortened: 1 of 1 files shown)


ehelms:
:fireworks: :man_dancing:

theforeman.org/pipelines/test/testKatello.groovy
```diff
@@ -106,6 +106,16 @@ pipeline {
...
                             dir('foreman') {
+                                sh "python script/filter-package-json.py"
+
+                                withRVM(["bundle exec ruby script/plugin_webpack_directories.rb > plugin_webpack.json"], ruby)
+                                script {
+                                    def plugin_webpack = readJSON file: 'plugin_webpack.json'
+                                    plugin_webpack['plugins'].each { plugin, config ->
+                                        sh "python script/filter-package-json.py --package-json ${config['root']}/package.json"
+                                    }
+                                }
```
ehelms, line 117: Do you think that this logic can be:

 a) wrapped up in a common groovy function
 b) wrapped up in a script that lives in foreman

The two use cases I am thinking of to extrapolate this awesome logic to:

 1) Using it in other plugin CI tests
 2) Using the same logic in Katello react tests that flow through GH actions

evgeni: Yes.

Groovy is easy, the other thing a bit more complicated but still possible.

ehelms: Another spot I was curious if any of this logic would benefit -- https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/scripts/test/test_develop.sh#L42-L45

evgeni: Yes, that file would benefit from being Groovy 😼

evgeni: But no, here we actually need more than the build deps (still no storybook), but I'd expect that one to break if we apply *exactly* the same logic

ehelms:
Since this provides a significant speed up I am happy to merge and then re-factor. As the author of the change, I'll let you have final say which order you want to go in.

evgeni:
Merge away

ehelms (approved):

PR #224, opened 2022-06-13
do not create empty gemsets for every job run

evgeni:


evgeni:
4 minutes → 1 minute:

image

ehelms (approved):
Without the cleanup step I think this should properly avoid any race conditions. I think the only downside here is there is effectively a cache per job per ruby that we will need to monitor.

ekohl:
Doesn't this mean different tests can leak into each other? What also comes to mind is pulling a too new gem, like we had in GH actions with Psych 4 where it wasn't being properly downgraded. We didn't suffer from it on Jenkins because we always create an empty gemset for every test.

ekohl:
> From some testing, and thinking on this with newer bundler, I think we could drop gemsets all together for the jobs themselves.

I had the same thought. Rather than changing the configureRVM method, perhaps it should simply stop using that.

evgeni:
> I had the same thought. Rather than changing the configureRVM method, perhaps it should simply stop using that.

Maybe. But this gave me a much smaller diff and keeps `bundler` up to date still.

evgeni:
> I think the only downside here is there is effectively a cache per job per ruby that we will need to monitor.

Yeah. We have a job that cleans up old gemsets if they are unused:
https://github.com/theforeman/foreman-infra/blob/9a5389d9578219c9d90b66f983028e4ca7d6b021/puppet/modules/slave/manifests/init.pp#L109

But now that these will be kept "used", I guess it won't clean up that much anymore.

evgeni:
this will be fixed by not using RVM alltogether.

PR #231, opened 2022-06-24
Add pipeline based Foreman PR tests split into three

ehelms:
Creates pipelines for Foreman PR tests and splits them into unique
jobs with unique contexts to run independently and show results
more indepdently to the user.

Trying out a different approach than what we have done previously by splitting out the various major types of tests so that each not only reports a status directly to the PR but can also be re-ran independently of the others in the case of a failure that is intermittent. My hope is with all the caching work around npm and gems on our nodes that we won't see a performance hit with this change but will gain a significant boost to developer understanding around test failures and this inherently creates some parallelization.


ehelms:
I debated that. Right now this info is only overlapped in two jobs and I
went this route for flexibility for each of the test scenarios.

ehelms:
I have moved the various mappings into a shared file.

ehelms:
Note also this only adds new -- so it could be merged and work alongside existing PR test structures to ensure each is working properly before turning off the old version of the job.

ehelms:
Additionally, if core maintainers would rather move any of this functionality into GH actions as tests let's talk about that as well.

@theforeman/core your input on the future of the test jobs would be appreciated to help in which direction to go

theforeman.org/pipelines/test/foreman/vars.groovy
```diff
@@ -0,0 +1,30 @@
...
+    '3.3-stable': [
+        'ruby': ['2.7', '2.5']
+    ],
+    '3.2-stable': [
+        'ruby': ['2.7', '2.5']
+    ],
+    '3.1-stable': [
+        'ruby': ['2.7', '2.5']
+    ]
+]
```
ekohl, line 29: I don't think the duplication is needed. In the integration tests you already only have Ruby 2.7 as the only axis anyway.

theforeman.org/pipelines/test/foreman/unit.groovy
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    stages {
+        stage('matrix') {
+            matrix {
+                agent { label 'fast' }
+                axes {
+                    axis {
+                        name 'ruby'
+                        values '2.7', '2.5'
```
ekohl, line 19: I wonder if you can use the `ruby_versions` variable instead.
```suggestion
                        values ruby_versions
```

But I vaguely recall you can't and need to resort to something like https://www.jenkins.io/blog/2019/12/02/matrix-building-with-scripted-pipeline/

theforeman.org/pipelines/test/foreman/integration.groovy
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    stages {
+        stage('matrix') {
+            matrix {
+                agent { label 'fast' }
+                axes {
+                    axis {
+                        name 'ruby'
+                        values '2.7'
```
ekohl, line 19: Would this work to always run only on the first entry (which should be the latest):
```suggestion
                        values ruby_versions[0]
```

Then the `when` part below is redundant. You can also use `foreman_unit_versions` and get rid of the duplication.

ekohl (commented):
Just a thought: can't we run `assets:precompile` as part of the integration tests? There is a big overlap between the two and it would save a lot of time: just a single `webpack compile` if you run `rake assets:precompile jenkins:integration`.

theforeman.org/pipelines/test/foreman/unit.groovy
ehelms: Correct, what you pass to this cannot be dynamic, it has to be static.

theforeman.org/pipelines/test/foreman/integration.groovy
ehelms: I could try this, but what you pass to this cannot be dynamic from my previous playing around

ehelms: I expected that when we started turning on Ruby 3.0 support, we were going to want to run integration tests against 3.0 for at least some period of time to ensure things worked in both environments.

ekohl: That's a fair point.

ehelms:
> Just a thought: can't we run `assets:precompile` as part of the integration tests? There is a big overlap between the two and it would save a lot of time: just a single `webpack compile` if you run `rake assets:precompile jenkins:integration`.

Implemented

ekohl (approved):
I'd be OK with testing this, but it's now two instead of 3. We should fix that up in the commit message.

PR #232, opened 2022-06-25
support source-less packages better

evgeni:


evgeni:
@upadhyeammit this should fix the other issues with building a debian package that doesn't really have a "source"

ehelms (approved):

PR #235, opened 2022-06-29
Switch foreman-selinux jobs to make for rhel8

ehelms:


ekohl (commented):
I wonder if this is really correct. We have EL7 builders so how reliable is this testing anyway? SELinux policy compilation does use host bits. IMHO we should ensure we have EL8 builders, at least for SELinux testing.

ekohl:
We dropped it so I opened https://github.com/theforeman/jenkins-jobs/pull/396 instead.

PR #237, opened 2022-07-01
use c8s as base system for pipelines

evgeni:


ehelms:
Is the available inventory the same as what we were using prior?

evgeni:
Which inventory are you talking about?

ehelms:
Hah, yea, that is a bit of an overloaded term. The available inventory of machines for our tests to run against over in ci.centos.org

evgeni:
Oh yeah, you mean the ones over at https://wiki.centos.org/QaWiki/CI/Duffy?
It's the same systems. Well, same rack/chassis, different blade.

I still need to do some more tests with my hacks.

evgeni:
okay, stuff's passing now!

ekohl (approved):
With this, can we drop EL7 support for forklift "host" support or are there other places where we need to support Ruby 2.0 and all those ancient things.

evgeni:
Upstream, this should be the last place.

Downstream, this is also used and I have no idea what the pipelines run on (cc @pcreech @zjhuntin @Odilhao)

PR #238, opened 2022-07-04
run pipelines with telemetry enabled

evgeni:


centos.org/pipelines/lib/foremanCentosJob.groovy
```diff
@@ -53,12 +52,25 @@ pipeline {
                 script {
                     extra_vars = buildExtraVars(extraVars: playBook['extraVars'])
 
+                    def otel_env = """
+                    export OTEL_EXPORTER_OTLP_HEADERS='${env.OTEL_EXPORTER_OTLP_HEADERS}'
+                    export OTEL_EXPORTER_OTLP_ENDPOINT='${env.OTEL_EXPORTER_OTLP_ENDPOINT}'
+                    export OTEL_TRACES_EXPORTER='${env.OTEL_TRACES_EXPORTER}'
+                    export TRACE_ID='${env.TRACE_ID}'
+                    export TRACEPARENT='${env.SPAN_ID}'
```
evgeni, line 60: these two don't seem to work as I thought they'd work :(

evgeni:
a run is visible here: https://ui.honeycomb.io/the-foreman/environments/test/result/t3ikdArFcHD/trace/e59tfNwg1SX?fields[]=c_name&fields[]=c_service.name&span=f865ce087d24a96f

ehelms (approved):
Data capture looks awesome!

evgeni:
One thing that's broken is the tree. The ansible results somehow don't end up as children of the `sh: ssh…` events.

But it's better than nothing, so I'll go ahead and merge it and we can try to fine tune that later

PR #241, opened 2022-07-13
Drop foreman develop PR upgrade job

ehelms:
The idea is to drop this as it's not as useful given it's not testing very much data and to explore a similar test using a more robust database.


ehelms:
I'll mention that.

PR #243, opened 2022-07-19
Remove test_develop_pr_core

ehelms:
The job was split out and replaced by foreman-pr-unit-test and
foreman-pr-integration-test.


ehelms:
Merge when comfortable with the new jobs. Testing them here -- https://github.com/theforeman/foreman/pull/9312

theforeman.org/yaml/jobs/test_develop_pr_core.yaml
```diff
@@ -1,42 +0,0 @@
...
-          values:
-            - postgresql-integrations
-            - postgresql
-      - axis:
-          type: label-expression
-          name: slave
-          values:
-            - fast
-    builders:
-      - shell: !include-raw: scripts/test/test_develop.sh
```
ekohl, line 35: Note to self: this is still used in `theforeman.org/yaml/builders/test_develop.yaml` so shouldn't be dropped yet.

ekohl (approved):

PR #244, opened 2022-07-20
Replace Katello PR test for Foreman with a pipeline based one

ehelms:


theforeman.org/scripts/test/test_katello.sh
```diff
@@ -1,79 +0,0 @@
-#!/bin/bash -ex
```
ekohl, line 1: This file is still used - probably in the source build step for the nightly package.

theforeman.org/pipelines/test/foreman/vars.groovy
```diff
@@ -1,15 +1,19 @@
 def foreman_integration_versions = [
     'develop': [
-        'ruby': ['2.7']
+        'ruby': ['2.7'],
+        'katello': ['master']
```
ekohl, line 4: We have the same table elsewhere, but transposed. I wonder if we can reuse this data somehow. Perhaps just store it in the same file so you only need to update a single file?
https://github.com/theforeman/jenkins-jobs/blob/036271a98c4952aa7b620d33081762a617d01901/theforeman.org/pipelines/test/testKatello.groovy#L1-L27

theforeman.org/scripts/test/test_katello.sh
ehelms: It's used by plugin testing for foreman-virt-who-configure, I meant to not trash it yet.

theforeman.org/pipelines/test/foreman/vars.groovy
ehelms: That crossed my mind while writing this up. I was thinking to introduce this first and then come back with a round of re-factor to look at the landscape of where this info is stored to centralize it in a way that makes sense.

ekohl (approved):

PR #248, opened 2022-08-01
Archive Gemfile.lock for hammer-cli

ehelms:


evgeni:
These PRs tell me, we should have a method that does bundle install and automatically archives the lock file.

And now back to my other ghosts.

ekohl (approved):
I think the ghost is right, but for now let's move this forward.

PR #251, opened 2022-08-04
Add Katello 4.6 pipeline

sjha4:
Adds 4.6 specific pipelines and vars. Haven't removed any older configs or pipelines but let me know if we should.


ekohl (changes requested):
We recently made some changes and haven't reflected that in tool_belt. Could you also add the mapping to `theforeman.org/pipelines/test/foreman/vars.groovy`? This means PRs to foreman.git will check out the proper Katello version for the stable branch.

theforeman.org/pipelines/test/foreman/vars.groovy
```diff
@@ -21,6 +25,9 @@ def foreman_unit_versions = [
     'develop': [
         'ruby': ['2.7']
     ],
+    '3.4-stable': [
+        'ruby': ['2.7']
+    ],
     '3.3-stable': [
```
sjha4, line 31: @ekohl..Thanks! Added 3.4 and 4.6 configs.

theforeman.org/pipelines/vars/katello/4.6.groovy
```diff
@@ -0,0 +1,16 @@
+def foreman_version = '3.4'
+def katello_version = '4.6'
+def foreman_el_releases = [
+    'el7',
```
ekohl, line 4: There won't be al EL7 release
```suggestion
```

theforeman.org/yaml/jobs/pipeline/katello-rpm-pipeline.yaml
```diff
@@ -21,4 +21,5 @@
       - '4.2'
       - '4.3'
       - '4.4'
-      - '4.5'
\ No newline at end of file
+      - '4.5'
+      - '4.6'
```
ekohl, line 25: Could you ensure a line ending at the end of the file?

ekohl (changes requested):

ekohl (approved):
Thanks!

PR #256, opened 2022-08-31
Archive Puppetfile.lock in the installer

ekohl:
This makes it easier to figure out which modules were used, without having to dive into the console logs.

A draft since this is untested now.


evgeni (approved):

evgeni:
moved to https://github.com/theforeman/foreman-installer/pull/945

ekohl:
This is the source release job, which we still have. Perhaps still worth doing, even though I haven't needed it since I wrote this almost 2 years ago.

PR #260, opened 2022-09-15
use the real inventory, not some hacked up file

evgeni:


centos.org/ansible/foreman.duffy.yml
```diff
@@ -0,0 +1,5 @@
+plugin: evgeni.duffy.inventory
```
evgeni, line 1: right now this inventory would return *all* our nodes, so parallel jobs would be "funny", I need to add a limit to the ansible run

ehelms: Is this what the PR is waiting on to move out of draft?

evgeni: correct.

evgeni: this is now "resolved" as I added code that should do what I want, but didn't test it yet

centos.org/pipelines/lib/foremanDuffyJob.groovy
```diff
@@ -30,6 +31,8 @@ pipeline {
         stage('Install Pipeline Requirements') {
             steps {
                 script {
+                    def duffy_session = readFile(file: 'jenkins-jobs/centos.org/ansible/duffy_session')
```
ekohl, line 34: What is this file?

evgeni: it gets written by provisionDuffy() to remember the session id, so that deprovisionDuffy() knows what box to drop

evgeni:
stuff is working now!

ekohl (approved):

PR #265, opened 2022-09-27
Align foreman rpm pipelines on push_foreman_rpms

ekohl:
This closer aligns the various pipelines and makes it easier to have more code reuse in a future version.

Replacement of https://github.com/theforeman/foreman-infra/pull/1487.

Please review carefully.


theforeman.org/pipelines/lib/release.groovy
```diff
@@ -1,6 +1,19 @@
...
+    for (distro in distros) {
+        if repo_type {
+            // Main foreman repository
+            if version == 'nightly' {
+                push_rpms_direct("foreman-${foreman_version}/${release}", "${foreman_version}/${release}")
+            } else {
+                push_rpms("foreman-${version}", 'releases', version, distro, keep_old_files)
+            }
+        } else {
+            push_rpms("foreman-${repo_type}-${version}", repo_type, version, distro, keep_old_files)
```
ehelms, line 14: If it makes it this far, I would think `repo_type` will be null/false and this line won't make as much sense?

ekohl: See, there was a mistake. This should be the other way around: here repo_type must be defined. Will update.

theforeman.org/pipelines/lib/release.groovy
```diff
@@ -1,6 +1,19 @@
...
-    push_rpms("foreman-${repo_type}-${version}", repo_type, version, distro)
+    keep_old_files = version != 'nightly'
+
+    for (distro in distros) {
+        if repo_type {
+            push_rpms("foreman-${repo_type}-${version}", repo_type, version, distro, keep_old_files)
+        } else {
+            // Main foreman repository
+            if version == 'nightly' {
+                push_rpms_direct("foreman-${foreman_version}/${distro}", "${foreman_version}/${distro}")
```
ekohl, line 11: We can align this if we can push to `releases/nightly/${distro}` instead of `nightly/${distro}`. Now `releases/nightly` is a symlink to `../nightly` but I'm leaning to swapping that around (`nightly` is a symlink to `releases/nightly`) so we can get rid of the if condition. In practice our `foreman-release` RPM already uses `yum.theforeman.org/releases/nightly` anyway.

evgeni:
Today, everything is using `push_foreman_staging_rpms(<pocket>, <version>, distro)` (in a loop over distros).
I guess inlining the loop would make sense, but the rest of this PR is obsolete, right?

ekohl:
Yes. I think https://github.com/theforeman/jenkins-jobs/pull/458 is a better approach to it, even if it's incomplete.

PR #270, opened 2022-09-30
add foreman and plugins duffy pipes

evgeni:


centos.org/jobs/foreman-pipelines.yml
```diff
@@ -118,3 +118,17 @@
...
+      - 'foreman-pipeline-{type}-{flavor}-{version}'
+    empty: ''
+    flavor:
+      - deb
+      - rpm
+    version:
+      - nightly
+    type:
+      - foreman
+      - plugins
```
evgeni, line 134: we could, for consistency, also use the `flavor` template for katello/luna/pulpcore? and hardcode flavor to `rpm`.

thoughts?

ekohl: I'd prefer that. That way we don't have jobs defined that will never run.

ekohl (approved):
I'm going to try this out and hopefully allow us to move forward.

PR #271, opened 2022-09-30
Perform scratch builds in parallel

ehelms:
I am wondering if we can do this now and get better performance and more obvious output.

### DIFF
--- a/theforeman.org/pipelines/test/rpm_packaging.groovy
+++ b/theforeman.org/pipelines/test/rpm_packaging.groovy
@@ -133,9 +133,14 @@ pipeline {
                 expression { packages_to_build }
             }
             steps {
+                for(int i = 0; i < packages_to_build.size(); i++) {
+                    def index = i
+                    packages[packages_to_build[index]] = {
+                        obal(action: "scratch", extraVars: ['build_package_download_logs': 'True', 'build_package_download_rpms': 'True'], packages: packages_to_build[index])
+                    }
+                }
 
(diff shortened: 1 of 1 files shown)


evgeni:
Can we limit the parallelization somehow? I don't think doing 20+ packages in parallel (a usual Rails upgrade) should yield good results

theforeman.org/pipelines/test/rpm_packaging.groovy
```diff
@@ -133,9 +133,14 @@ pipeline {
                 expression { packages_to_build }
             }
             steps {
+                for(int i = 0; i < packages_to_build.size(); i++) {
+                    def index = i
+                    packages[packages_to_build[index]] = {
```
ekohl, line 138: I'd prefer:
```suggestion
                packages = [:]
                packages_to_build.each { package ->
                    packages[package] = {
```

ekohl (commented):
Koji does have queuing so perhaps that covers it enough? In theory it also means we could add a builder on days where we'd have many build jobs, though in practice that's unlikely to be something we need to optimize for.

Or are you worried about the Jenkins workers. In that case we could also look into implementing the concurrency in obal, given Ansible can also handle some concurrency.

evgeni:
I'm mostly worried about Koji starving itself.
When you trigger a build it contains multiple steps (for scratch: rebuild srpm, build arch, for release: rebuild srpm, build arch, tag, regen all affected repos).

This PR only touches scratch, so it's not *that* dangerous (srpm and arch don't do too much locking), so we might just wait for the individual results longer. But if we apply the same logic to release, I suspect to see the overall throughput to go down as jobs wait on repo regens etc.

PR #280, opened 2022-10-11
don't put individual playbooks in own stages

evgeni:
Jenkins doesn't cope well with parallel stages inside stages when we dynamically generate the list of steps.

If we'd use just `parallel` or `matrix` with a static set of steps, it would work, but we generate things on the fly and this confuses the Blue Ocean view, resulting in only the first stage to be present in the view and thus making it impossible to jump to the failed step if there are any.

### DIFF
--- a/centos.org/pipelines/lib/foremanDuffyJob.groovy
+++ b/centos.org/pipelines/lib/foremanDuffyJob.groovy
@@ -90,9 +90,7 @@ pipeline {
 
                                 duffy_scp_in('otel_env', 'otel_env', boxname, './')
                                 for(playbook in playbooks) {
-                                    stage(playbook) {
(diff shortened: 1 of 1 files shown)


centos.org/pipelines/lib/foremanDuffyJob.groovy
```diff
@@ -90,9 +90,7 @@ pipeline {
 
                                 duffy_scp_in('otel_env', 'otel_env', boxname, './')
                                 for(playbook in playbooks) {
```
ekohl, line 92: Thinking out loud: If we don't have a stage for it, does it still make sense to run it individually? I guess it's seen in the command so still some added value.

The reason I ask is that we can just run a single large pipeline and remove complexity.

ekohl (approved):

PR #282, opened 2022-10-18
Also deploy Puppet code to the new puppet Infra

ekohl:
This should pass, but I haven't tried it yet. This will help me to iterate on it.


ekohl:
I needed to install g10k manually (wasn't part of Puppet) but otherwise this now passed.

PR #284, opened 2022-10-20
split changelog data generation and injection

evgeni:
this allows to re-use the injection method for other things

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -287,7 +287,7 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
 
         if (pull_request || version == 'nightly') {
             dir("${project}-${package_version}") {
-                add_debian_changelog(os, package_version, repoowner, last_commit)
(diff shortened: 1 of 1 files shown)


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -1,3 +1,11 @@
+def debian_release_to_version = [
+    'buster': 'deb10',
+    'bullseye': 'deb11',
+    'bookworm': 'deb12',
```
evgeni, line 4: should this say `debian` instead of `deb`?

ehelms (approved):

theforeman.org/pipelines/lib/packaging.groovy
ehelms: That feels cleaner, and if you make that change and want to merge go for it.

evgeni: Will do (tomorrow)

evgeni: done, and added 13 and 14 codenames as they are already known

PR #296, opened 2022-11-18
disable telemetry for now

evgeni:


evgeni:
this is a workaround for
```
ERROR! Unexpected Exception, this is probably a bug: The 'resolvelib<0.9.0,>=0.5.3' distribution was not found and is required by ansible-core
```

evgeni:
for tracking, this is https://bugzilla.redhat.com/show_bug.cgi?id=2143974

PR #303, opened 2023-01-03
Add a smart-proxy-stable pipeline

ekohl:
This is intended to replace the `test_proxy_X_Y` jobs with pipelines. This is intended to make the naming more consistent and branching easier. It also works towards phasing out the legacy jobs. Right now it doesn't remove the old ones so I can easily experiment with it.

Also includes a commit that externalizes the testing to a script so it's easy to run it locally.

### DIFF
--- a/theforeman.org/pipelines/lib/stable/smart-proxy.groovy
+++ b/theforeman.org/pipelines/lib/stable/smart-proxy.groovy
@@ -1,0 +1,13 @@
+properties([
+    parameters([
+        gitParameter(branch: '',
+                     defaultValue: git_branch,
+                     description: '',
+                     name: 'BRANCH',
+                     quickFilterEnabled: false,
+                     selectedValue: 'NONE',
+                     sortMode: 'NONE',
+                     tagFilter: '*',
+                     type: 'PT_BRANCH')
+    ])
+])
--- a/theforeman.org/yaml/jobs/pipeline/foreman-stable-pipelines.yaml
+++ b/theforeman.org/yaml/jobs/pipeline/foreman-stable-pipelines.yaml
@@ -1,0 +1,21 @@
+- job-template:
+    name: 'smart-proxy-{version}-stable-test'
+    project-type: pipeline
+    sandbox: true
+    properties:
+      - github:
+          url: https://github.com/theforeman/smart-proxy
+    dsl:
+      !include-raw:
+        - 'pipelines/vars/foreman/{version}.groovy'
+        - 'pipelines/lib/stable/smart-proxy.groovy{empty}'
+        - 'pipelines/test/smart-proxy.groovy{empty}'
+        - 'pipelines/lib/rvm.groovy{empty}'
+
+- project:
+    name: foreman-stable
+    jobs:
+      - 'smart-proxy-{version}-stable-test'
+    empty: ''
+    version:
+      !include: ../../includes/foreman_versions.yaml.inc
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Useful resources:
 | Nightly Package Builder | {git-repo}-{git-branch}-package-release                | foreman-develop-package-release | hammer-cli-katello-master-package-release |
 | CI pipeline             | {repository}-{environment}-{optional-concern}-pipeline | foreman-nightly-rpm-pipeline    | foreman-nightly-deb-pipeline              |
 | Pull Request testing    | {git-repo}-{optional-concern}-pr-test                  | katello-pr-test                 | foreman-packaging-rpm-pr-test             |
+| Branch testing          | {git-repo}-{git-branch}-test                           | foreman-3.5-stable-test         | smart-proxy-3.5-stable-test               |
 
 ## Job configurations
 
--- a/branch-foreman
+++ b/branch-foreman
@@ -8,7 +8,8 @@ if [[ -z $FOREMAN_VERSION ]] || [[ -z $KATELLO_VERSION ]] ; then
 	exit 1
 fi
 
-sed "/foreman_version/ s/nightly/${FOREMAN_VERSION}/" theforeman.org/pipelines/vars/foreman/nightly.groovy > theforeman.org/pipelines/vars/foreman/${FOREMAN_VERSION}.groovy
+sed "/foreman_version/ s/nightly/${FOREMAN_VERSION}/ ; /git_branch/ s/develop/\${foreman_version}-stable/" \
+	theforeman.org/pipelines/vars/foreman/nightly.groovy > theforeman.org/pipelines/vars/foreman/${FOREMAN_VERSION}.groovy
 
 sed -e "/foreman_version/ s/nightly/${FOREMAN_VERSION}/" \
 	-e "/katello_version/ s/nightly/${KATELLO_VERSION}/" \
--- a/theforeman.org/pipelines/test/smart-proxy.groovy
+++ b/theforeman.org/pipelines/test/smart-proxy.groovy
@@ -1,3 +1,11 @@
+def git_checkout = {
+    if (env.ghprbPullId) {
+        ghprb_git_checkout()
+    } else {
+        git branch: "${params.BRANCH}", url: 'https://github.com/theforeman/smart-proxy'
+    }
+}
+
 pipeline {
     agent none
     options {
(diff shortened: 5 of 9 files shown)


theforeman.org/pipelines/vars/smart-proxy/3.3.groovy
```diff
@@ -0,0 +1 @@
+git_branch = '3.3-stable'
```
ekohl, line 1: These files are really trivial now. I'm still debating what's best. I imagine that we should either find a way to avoid this or expand on this where this file also contains the variables like Ruby versions.

theforeman.org/pipelines/test/smart-proxy-stable.groovy
```diff
@@ -0,0 +1,113 @@
...
+            }
+        }
+
+        stage('Test') {
+            matrix {
+                agent any
+                axes {
+                    axis {
+                        name 'ruby'
+                        values '2.7'
```
ekohl, line 63: This has to be static: https://issues.jenkins.io/browse/JENKINS-62127

theforeman.org/pipelines/test/smart-proxy-stable.groovy
```diff
@@ -0,0 +1,113 @@
...
+                    }
+                }
+                stage('Install dependencies') {
+                    steps {
+                        withRVM(['bundle install'], rubocop_ruby)
+                    }
+                }
+                stage('Run Rubocop') {
+                    steps {
+                        withRVM(['bundle exec rubocop --format progress --out rubocop.log --format progress'], rubocop_ruby)
```
ekohl, line 40: I copied this from the other pipeline, but while working on it I noticed this has `--format progress` twice.

theforeman.org/pipelines/test/smart-proxy-stable.groovy
```diff
@@ -0,0 +1,113 @@
...
+        stage('Rubocop') {
+            agent any
+            environment {
+                BUNDLE_WITHOUT = 'bmc:development:dhcp_isc_inotify:dhcp_isc_kqueue:journald:krb5:libvirt:puppetca_token_whitelisting:realm_freeipa:windows'
+            }
+
+            stages {
+                stage('Setup Git Repos') {
+                    steps {
+                        git(url: git_url, branch: git_branch)
```
ekohl, line 25: This is the only (real) difference to the PR testing pipeline. I'm wondering if we should externalize the git checkout step to a function.

ekohl (commented):
@ehelms @evgeni I'd appreciate a look at this. My goal is to replace the old `test_X_Y` and `test_proxy_X_Y` jobs. That gets rid of legacy jobs and makes branching (and cleanup of old branches) easier. 

https://github.com/theforeman/theforeman-rel-eng/blob/master/verify_green_jobs should also be updated once the jobs are replaced.

ekohl:
I ran this manually and it passes: https://ci.theforeman.org/job/smart-proxy-3.5-stable-test/1/ (though later today CI will remove the untracked file again).

theforeman.org/pipelines/vars/smart-proxy/3.3.groovy
evgeni: can't we just do `git_branch = "{version}-stable"` in some more central place?

ekohl: Any thoughts where? I thought about https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/vars/foreman/3.5.groovy but wonder if it makes sense.

evgeni: foreman follows the same branching structure, so if you also want to replace the foreman stable_X_Y jobs (and I guess we do), that's a good place?

ekohl: > if you also want to replace the foreman stable_X_Y jobs

That is indeed the intent.

theforeman.org/pipelines/test/smart-proxy-stable.groovy
ekohl: I now applied this in the first commit.

theforeman.org/pipelines/lib/stable/smart-proxy.groovy
```diff
@@ -0,0 +1,3 @@
+def git_checkout() {
+    git(url: 'https://github.com/theforeman/smart-proxy', branch: git_branch)
```
ekohl, line 2: @evgeni I can't quite figure out how I would get `git_branch` from the top level into the function. Adding a parameter like this doesn't work either. Any suggestion?
```suggestion
def git_checkout(git_branch=git_branch) {
    git(url: 'https://github.com/theforeman/smart-proxy', branch: git_branch)
```

evgeni: I'll try to think tomorrow ;)

theforeman.org/yaml/jobs/pipeline/foreman-stable-pipelines.yaml
```diff
@@ -0,0 +1,22 @@
+- job-template:
```
ekohl, line 1: This probably needs triggers to avoid having to build at least once manually.

theforeman.org/pipelines/lib/stable/smart-proxy.groovy
evgeni: I think the only way to do it is to define `git_checkout` with a `git_branch` parameter and then actually pass that from the caller:


```groovy
def git_checkout(git_branch) {
…
}
…
                stage('Setup Git Repos') {
                    steps {
                        git_checkout(git_branch)
                    }
                }
```

ekohl: The problem with that is that `git_branch` is undefined for PR builds.

https://stackoverflow.com/questions/6305910/how-do-i-create-and-access-the-global-variables-in-groovy suggests that if you use `git_branch = '..'` instead of `def git_branch = '..'` it may work. I'll try that first.

ekohl: That works so I think this PR is now ready.

ehelms: Relying on a "global" to exist is generally a bad pattern and can lead to a variety of issues from my experience. Reading back, I interpreted this as -- sometimes git_branch will be defined and sometimes it will be effectively `null` ?

ehelms: > The problem with that is that `git_branch` is undefined for PR builds.

Why is this true? I would have thought in a PR test the git_branch would be defined as the either the target branch or the branch of the PR itself from the users fork.

ehelms: What about something like:

```
def git_checkout(git_branch) {
    if (git_branch) {
           git(url: 'https://github.com/theforeman/smart-proxy', branch: git_branch)
    } else {
           ghprb_git_checkout()
    }
}
```

You could even make debugging a bit easier:

```
def git_checkout(git_branch) {
    if (git_branch) {
           git(url: 'https://github.com/theforeman/smart-proxy', branch: git_branch)
    } else if (ghprbTargetBranch) {
           ghprb_git_checkout()
    } else {
          error("No git_branch or PR target branch specified")
    }
}
```

ekohl: I suppose I could use `git_branch = env.ghprbTargetBranch` or `git_branch = null`, but it feels wrong if it's not going to be used anyway. You would end up with:

```groovy
def git_checkout(git_branch) = {
    ghprb_git_checkout()
}
```
I also thought about [currying](https://groovy-lang.org/closures.html#_currying) but you can't do that outside of script blocks (since you can't invoke function).

I'm beginning to think we need libraries. However, I just found https://plugins.jenkins.io/git-parameter/ which we have installed, so I'm going to play with that as well.

ekohl: How about this version?

ehelms (approved):
If you tested and this works, I'm good with it.

ekohl:
I did test it and it worked. Not sure about the trigger for the first run, but that's a problem we also have in a few other jobs.

@evgeni any last comments?

evgeni (approved):
go for it

ekohl:
Follow up work:
* https://github.com/theforeman/theforeman-rel-eng/pull/181
* https://github.com/theforeman/tool_belt/pull/500
* https://github.com/theforeman/jenkins-jobs/pull/306

PR #304, opened 2023-01-03
Enhance configureRVM to install a specific bundler

ekohl:
Bundler 2.4 came out and dropped support for Ruby 2.5, which we still test with for Puppet 6 support.


theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -13,9 +13,13 @@ def gemset(name = null) {
...
 
-def configureRVM(ruby, name = '') {
+def configureRVM(ruby, name = '', bundler_version = null) {
     emptyGemset(ruby, name)
-    withRVM(["gem install bundler --no-document"], ruby, name)
+    if (bundler_version) {
+        withRVM(["gem install bundler -v '${bundler_version}' --no-document"], ruby, name)
+    } else {
+        withRVM(["gem install bundler --no-document"], ruby, name)
+    }
```
evgeni, line 22: the alternative would be to fall back `bundler_version` to something like `>= 0`, but I don't mind either way

evgeni (approved):

theforeman.org/pipelines/lib/rvm.groovy
ekohl: I thought this was easier to read in the job output

PR #308, opened 2023-01-27
Adding Config for Pulpcore 3.22

Odilhao:


theforeman.org/yaml/includes/pulpcore_versions.yaml.inc
```diff
@@ -1,4 +1,5 @@
 - '3.16'
 - '3.17'
 - '3.18'
-- '3.21'
\ No newline at end of file
+- '3.21'
+- '3.22'
```
ekohl, line 5: Please ensure a line end is here.

Odilhao: Ack

ekohl: It still shows the symbol that there's no line ending.

Odilhao: Had to use vim, lol

ekohl: I don't know why you keep marking it as resolved, but it's still not there.

ekohl (approved):

PR #309, opened 2023-02-10
Add Katello 4.8 pipeline

lfu:


chris1984:
@ekohl how does this look?

ekohl (changes requested):
I would like to do complete branching together with Foreman. I already wrote https://github.com/theforeman/jenkins-jobs/blob/master/branch-foreman and my goal is that we do that. Branching Katello without Foreman is rather pointless.

I actually already submitted https://github.com/theforeman/tool_belt/commit/579c5c790c3a6afa85f1e71fd9606e383a58054a so it looks like the branching procedure for Katello was posted based on some outdated commit.

Looking at this PR I have at least missed adding `foreman_unit_versions` to `theforeman.org/pipelines/test/foreman/vars.groovy`, other than that this PR matches what the script does and asks you to do. So it's still good that this was submitted because it exposed an issue.

Would you mind expanding this PR to include Foreman jobs pr the script? Bonus points for extending the script with the `foreman_unit_versions` snippet and updating the procedure on Discourse.

chris1984:
@ekohl fixed the script and used it to make the changes requested. Will update Discourse as well.

ekohl (approved):
Thanks! I'll merge this when we actually branch next week.

PR #315, opened 2023-03-02
trigger deb-test-plugins after deb has passed

evgeni:
for rpm we do something similar with foreman → katello → luna but for deb we don't trigger plugins regularly

### DIFF
--- a/theforeman.org/yaml/jobs/pipeline/foreman-release-pipelines.yaml
+++ b/theforeman.org/yaml/jobs/pipeline/foreman-release-pipelines.yaml
@@ -67,6 +67,11 @@
     name: 'foreman-plugins-{version}-deb-test-pipeline'
     project-type: pipeline
     sandbox: true
+    triggers:
+      - reverse:
+          jobs:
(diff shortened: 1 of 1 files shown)


theforeman.org/yaml/jobs/pipeline/foreman-release-pipelines.yaml
```diff
@@ -67,6 +67,11 @@
     name: 'foreman-plugins-{version}-deb-test-pipeline'
     project-type: pipeline
     sandbox: true
+    triggers:
+      - reverse:
+          jobs:
+            - foreman-{version}-deb-pipeline
```
evgeni, line 73: this will also create triggers for `foreman-3.6-deb-pipeline` etc, which don't exist, but I think this is fine and makes the smallest change to achieve the result

ehelms (approved):

evgeni:
this worked just fine: https://ci.theforeman.org/job/foreman-plugins-nightly-deb-test-pipeline/13/

the non-nightly jobs now have a warning when you edit them, but who cares XD

PR #317, opened 2023-03-07
Add Ruby 3 testing to Smart Proxy pipelines

archanaserver:
Add Ruby 3 testing to Smart Proxy pipelines

**Requires:** 

This PR has dependecies on `smart-proxy` core and it's plugins PR for ruby 3 upgrade

### DIFF
--- a/theforeman.org/pipelines/release/source/smart-proxy.groovy
+++ b/theforeman.org/pipelines/release/source/smart-proxy.groovy
@@ -14,9 +14,9 @@ pipeline {
                 git url: git_url, branch: git_ref
             }
         }
-        stage("test ruby-2.7") {
+        stage("test ruby") {
             steps {
-                run_test(ruby: '2.7')
+                run_test(ruby: '2.7', '3.0', '3.1')
             }
         }
         stage('Build and Archive Source') {
--- a/theforeman.org/pipelines/test/smart-proxy.groovy
+++ b/theforeman.org/pipelines/test/smart-proxy.groovy
@@ -67,7 +67,12 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '2.7'
+                        values '3.1'
+                    }
+                }
+                when {
+                    expression {
+                        ruby_versions.contains(env.ruby)
                     }
                 }
                 environment {
@@ -78,6 +83,10 @@ pipeline {
                         steps {
                             script {
                                 git_checkout()
+                            }
+
+                            dir('foreman') {
+                                git url: "https://github.com/theforeman/foreman", branch: develop, poll: false
                             }
                         }
                     }
--- a/theforeman.org/pipelines/test/smart-proxy.groovy
+++ b/theforeman.org/pipelines/test/smart-proxy.groovy
@@ -67,12 +67,7 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '3.1'
-                    }
-                }
-                when {
-                    expression {
-                        ruby_versions.contains(env.ruby)
+                        values '2.7', '3.0', '3.1'
                     }
                 }
                 environment {
@@ -83,10 +78,6 @@ pipeline {
                         steps {
                             script {
                                 git_checkout()
-                            }
-
-                            dir('foreman') {
-                                git url: "https://github.com/theforeman/foreman", branch: develop, poll: false
                             }
                         }
                     }


theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -67,7 +67,12 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '2.7'
+                        values '3.1'
```
ekohl, line 70: We want to test a matrix, namely 2.7, 3.0 and 3.1. This is because we still want to make sure Ruby 2.7 continues working. There may also be differences between 3.0 and 3.1 so we want to verify both.
```suggestion
                        values '2.7', '3.0', '3.1'
```

theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -79,6 +84,10 @@ pipeline {
                             script {
                                 git_checkout()
                             }
+
+                            dir('foreman') {
+                                git url: "https://github.com/theforeman/foreman", branch: develop, poll: false
+                            }
```
ekohl, line 90: The Smart Proxy shouldn't include Foreman
```suggestion
```

ekohl (changes requested):

theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -79,6 +84,10 @@ pipeline {
                             script {
                                 git_checkout()
                             }
+
+                            dir('smart-proxy') {
+                                git url: "https://github.com/theforeman/smart-proxy", branch: develop, poll: false
+                            }
```
ekohl, line 90: I still fail to see the need for this. AFAIK plugins list smart-proxy as a dependency in `Gemfile` so it shouldn't be needed to check this out. In case of the Smart Proxy itself you just end up with a duplicate directory.
```suggestion
```

ekohl (changes requested):

theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -67,7 +67,12 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '2.7'
+                        values '2.7', '3.0', '3.1'
+                    }
+                }
+                when {
+                    expression {
+                        ruby_versions.contains(env.ruby)
```
ekohl, line 75: I get the impression you're copying this from `foreman.groovy`, but there it's defined in another file: https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/foreman/vars.groovy

Then it's set at the start of the file:
https://github.com/theforeman/jenkins-jobs/blob/39e5d043a8ab292206210577cf3d0fb572ed387a/theforeman.org/pipelines/test/foreman/unit.groovy#L2

In this case you don't, so it will fail. I think it's safe to leave out the `when` expression.

ekohl (changes requested):

theforeman.org/pipelines/test/smart-proxy.groovy
ehelms: I think this will result in running release branches against Ruby 3 versions as well. We can handle this by adapting the logic from Foreman's tests:

 * Add a when condition: https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/foreman/unit.groovy#L23-L27
 * https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/foreman/unit.groovy#L2
 * Update https://github.com/theforeman/jenkins-jobs/blob/91b6fcf3cf7b11b15964eba9228d05431d9b741a/theforeman.org/pipelines/test/foreman/vars.groovy#L32

ekohl (commented):
I think this is a good start, but I'd be more specific in the description (commit message, PR description). Something like "Add Ruby 3 testing to Smart Proxy pipelines" is closer to what it ends up being.

After we merge this it's good to kick off tests for all tests that use these pipelines to see if they still pass. I'd highly recommend using [hub](https://github.com/github/hub) if you want to create a bulk PRs. Let me know if you need help with that.

theforeman.org/pipelines/test/smart-proxy.groovy
ekohl: We had an offline discussion, but I recommended to just proceed with the simple version. In  https://github.com/theforeman/smart-proxy/commit/a921761d7f8ba5f2aadb793872cc17b5ba1b062f (https://projects.theforeman.org/issues/35510) the tests started to pass. I now see that was only in 3.5.0 and we still support 3.4 now.

Though I do think it's easier if we add `webrick` to Gemfile in 3.4-stable. That way we do pull it in for dev setups, but don't make any packaging change. Then I'd expect it to just pass and the CI code is MUCH easier.

ehelms: That's fine, it's likely this will have to be re-visited on the backend where we start dropping Ruby 2.7 support but our release branches still need it.

ekohl: Yes, but I'm fine with making that a "future us"-problem ;)

archanaserver:
@ekohl I've changed the description.

> I think this is a good start, but I'd be more specific in the description (commit message, PR description). Something like "Add Ruby 3 testing to Smart Proxy pipelines" is closer to what it ends up being.
> 
Yes it would be really helpful and more clear for me.
> After we merge this it's good to kick off tests for all tests that use these pipelines to see if they still pass. I'd highly recommend using [hub](https://github.com/github/hub) if you want to create a bulk PRs. Let me know if you need help with that.

theforeman.org/pipelines/release/source/smart-proxy.groovy
```diff
@@ -14,9 +14,9 @@ pipeline {
                 git url: git_url, branch: git_ref
             }
         }
-        stage("test ruby-2.7") {
+        stage("test ruby") {
```
archanaserver, line 17: i've removed the 2.7 text from here because now we don't have a single ruby version to point to.

ehelms:
LGTM -- @ekohl ?

ekohl (approved):
I think we can merge this and see where the pieces land.

ekohl:
And I just realized it was a PR with multiple commits and I didn't squash. Something to improve next time.

archanaserver:
> And I just realized it was a PR with multiple commits and I didn't squash. Something to improve next time.

Sure, thanks @ekohl !

theforeman.org/pipelines/release/source/smart-proxy.groovy
```diff
@@ -14,9 +14,9 @@ pipeline {
...
             steps {
-                run_test(ruby: '2.7')
+                run_test(ruby: '2.7', '3.0', '3.1')
```
ekohl, line 19: We missed in the review, but this is not valid. https://ci.theforeman.org/job/smart-proxy-develop-source-release/175/console now has:
> java.lang.NoSuchMethodError: No such DSL method 'run_test' found among steps

This is because there's no method that fits the method signature.

What should have been done was to convert the stage to a matrix, like in the other file.

archanaserver: @ekohl https://github.com/theforeman/jenkins-jobs/pull/324

PR #318, opened 2023-03-09
Remove database parameter to testing & run integration tests on test_develop

ekohl:
The plugin tests were failing because database was not set for some reason I didn't investigate:

```
13:02:33 + sed 's/^test:/development:/; s/database:.*/database: test_plugin_matrix-0-dev/' /home/jenkins/.db.yaml
13:02:33 sed: can't read /home/jenkins/.db.yaml: No such file or directory
```

If we ever pass `database`, the only value we pass is postgresql (the only DB Foreman 2+ supports) so we can drop it.

Also runs integration tests on test_develop. Since c5c1bd6c8e635f87eb4c447f66e64b49ff38b3c7 this code wasn't triggered anymore, but it's actually a good idea to run integration tests.

### DIFF
--- a/theforeman.org/scripts/test/test_develop.sh
+++ b/theforeman.org/scripts/test/test_develop.sh
@@ -25,11 +25,11 @@ bundle exec rake rubocop
 
 # Database environment
 (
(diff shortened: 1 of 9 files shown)


ehelms (approved):

ekohl:
Rebased to resolve a merge conflict.

PR #328, opened 2023-04-13
Remove Katello versions eariler than 4.7 per the release process

chris1984:
Per this step in the release process:

Remove references to the oldest Katello version from [testKatello.groovy 1](https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/test/testKatello.groovy) and [katello-pipelines.yml 1](https://github.com/theforeman/jenkins-jobs/blob/master/centos.org/jobs/katello-pipelines.yml).

This PR covers this

### DIFF
--- a/centos.org/jobs/katello-pipelines.yml
+++ b/centos.org/jobs/katello-pipelines.yml
@@ -5,7 +5,6 @@
       - 'foreman-pipeline-{type}-{flavor}-{version}'
     empty: ''
     version:
-      - '4.6'
       - '4.7'
       - '4.8'
       - 'nightly'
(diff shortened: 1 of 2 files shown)


chris1984:
@ekohl do I need to update anything on this to fix the failure?

evgeni:
Can you rebase on latest master? This should fix it.

ekohl:
@evgeni it doesn't. See https://github.com/theforeman/jenkins-jobs/pull/329. I think we should pin to 4.3.0, like we did with foreman-infra.

evgeni:
Ooh, I thought we did pin it here

PR #331, opened 2023-04-26
don't set per-build-tag on the hammer test jobs

evgeni:


evgeni:
https://ci.theforeman.org/job/test_hammer_cli_foreman_pull_request/606 is broken because it tries to tag (but can't due to miss-config), but https://ci.theforeman.org/job/test_hammer_cli_foreman_pull_request/607/ passed fine with this PR applied

PR #333, opened 2023-05-02
pin jjb to <5

evgeni:
Fixes: 440f7de76be6fafb553f908e2aef2e9b72a30a0d

### DIFF
--- a/theforeman.org/pipelines/infra/updateJobs.groovy
+++ b/theforeman.org/pipelines/infra/updateJobs.groovy
@@ -14,7 +14,7 @@ pipeline {
                 ])
 
                 virtEnv('jjb-venv', 'pip install pbr')
-                virtEnv('jjb-venv', 'pip install jenkins-job-builder')
+                virtEnv('jjb-venv', 'pip install -r requirements.txt')
             }
         }
 


theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -14,7 +14,7 @@ pipeline {
                 ])
 
                 virtEnv('jjb-venv', 'pip install pbr')
-                virtEnv('jjb-venv', 'pip install jenkins-job-builder')
+                virtEnv('jjb-venv', 'pip install jenkins-job-builder<5')
```
evgeni, line 17: you could argue, this should be a `-r requirements.txt` instead?

ekohl (approved):

theforeman.org/pipelines/infra/updateJobs.groovy
ekohl: Indeed. I'd prefer that.

theforeman.org/pipelines/infra/updateJobs.groovy
```diff
@@ -14,7 +14,7 @@ pipeline {
                 ])
 
                 virtEnv('jjb-venv', 'pip install pbr')
```
ekohl, line 16: Is this still needed? I imagine pip should be able to do this.

evgeni: you would, right? (it didn't last time we tried)

PR #334, opened 2023-05-02
Refs #36345 - Drop Puppet 6 support

ekohl:


evgeni:
Line 43/44 also references Ruby 2.5 which we now can drop/adjust?

evgeni:
Damn, I shouldn't review while getting the kids to sleep. Forgetting half the things.

At least https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/pipelines/release/source/foreman-installer.groovy also references p6/r2.5

ekohl:
I'm not sure why I didn't fix that, because I had the file open.

evgeni (approved):
There is also `theforeman.org/pipelines/test/kafo.groovy` which could use some pruning, but it's not Foreman Installer *per se*.

PR #338, opened 2023-05-11
Reuse test_plugin.sh in plugin PR testing

ekohl:
This reuses the test_plugin.sh script in plugin PR testing to avoid having to duplicate steps.


ekohl:
Perhaps it was done this way because concatting multiple shell fragments isn't supported?
```
 jenkins_jobs.errors.JenkinsJobsException: JOB_NAME parameter missing to format #!/bin/bash -ex
```

PR #339, opened 2023-05-11
Name multiple SCMs

ekohl:
This uses the multi-scms plugin[1] to define multiple SCMs for a single job.

[1]: https://plugins.jenkins.io/multiple-scms/


ekohl:
I've decided to revert the change and not use the multi-scms plugin.

PR #346, opened 2023-07-22
Drop smart_proxy_dynflow PR test

ehelms:
I was reviewing https://github.com/theforeman/smart_proxy_dynflow/pull/123 and noticed at this point the smart_proxy_dynflow repository runs the same set of tests in Github Actions and Jenkins. I feel we can drop this duplication in Jenkins.

### DIFF
--- a/theforeman.org/yaml/jobs/smart-proxy-plugin.yaml
+++ b/theforeman.org/yaml/jobs/smart-proxy-plugin.yaml
@@ -45,7 +45,6 @@
       - dns_infoblox
       - discovery:
           branch: develop # have to be the special child
-      - dynflow
       - monitoring
       - omaha
       - openscap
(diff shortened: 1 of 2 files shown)


ehelms:
@ekohl @evgeni Any objections to this change you can think of?

ekohl:
The only thing I can think of is the weekly test we now run, but I'm not sure that adds any real value right now. https://ci.theforeman.org/job/smart-proxy-dynflow-test/

ehelms:
I'm not convinced it does either, and it has low visibility. We could opt to run the Github Actions on a schedule if we wanted to achieve this same idea.

ehelms:
Added a drop for the weekly test

ehelms:
@ekohl give this another look?

ekohl (approved):

PR #349, opened 2023-08-15
Add Copr nightly test pipeline

ehelms:


ehelms:
I am using these as a playground, we do not necessarily need independent jobs in the end. I've expressed this here the options as I see them:

https://github.com/theforeman/forklift/pull/1700#issuecomment-1697466649

ehelms:
This will be adapted based upon the approach settled with the client -- https://github.com/theforeman/jenkins-jobs/pull/351

ehelms:
Proposed order of operations:

 1. Merge this change tat keeps the regular Koji pipeline running but adds the switch to stagingyum
 2. Swap to stagingyum in our CI job (https://github.com/theforeman/jenkins-jobs/pull/365)
 3. Swap to stagingyum in Forklift pipelines (https://github.com/theforeman/forklift/pull/1717)

theforeman.org/pipelines/release/pipelines/foreman-rpm.groovy
```diff
@@ -9,42 +9,114 @@ pipeline {
...
+                            deleteDir()
+                        }
+                    }
+                }
+                stage('staging-install-test') {
+                    agent any
+
+                    steps {
+                        script {
+                            runDuffyPipeline('foreman-rpm', foreman_version)
```
evgeni, line 107: should/could this pass `staging_source` somehow, so we don't need https://github.com/theforeman/forklift/pull/1717?

evgeni (approved):

theforeman.org/pipelines/release/pipelines/foreman-rpm.groovy
evgeni: mhh, `runDuffyPipeline` doesn't allow us to pass arbitrary parameters to the job, so probably not so easy.

ehelms: I reached the same conclusion. However, forklift should match CI if someone goes to run things so I think it's OK.

PR #351, opened 2023-08-30
Adapt client release pipeline to use Koji or Staging repositories

ehelms:
I am opting to go with a single pipeline definition but put a switch in place to allow us to flip from Koji to Copr (or back again if there are issues). I am starting with client as it's the least volatile and can prove out the design. 

This should allow us to merge this as is, and then flip the switch when we are ready with a simple code update (or easy local testing).


ehelms:
This can be merged as-is now since it does not switch to staging yet but does put in place the ability to switch and allows us to run the Koji based method still but with these re-factorings.

To switch to stage I will still need:

 * https://github.com/theforeman/theforeman-rel-eng/pull/262
 * https://github.com/theforeman/foreman-infra/pull/1875

PR #360, opened 2023-10-02
Remove jobs for foreman_openscap

ofedoren:
There is no longer use in jenkins job: https://github.com/theforeman/foreman_openscap/pull/545#issuecomment-1737403073

Leaving job for smart_proxy_openscap since we don't have it covered via Github Actions.


ofedoren:
Closing due to https://projects.theforeman.org/issues/36804.

PR #368, opened 2023-10-31
Build nightly Copr packages

ehelms:


theforeman.org/pipelines/release/foreman-x-develop-release.groovy
```diff
@@ -42,6 +42,25 @@ pipeline {
                                 setup_obal()
                             }
                         }
+                        stage('copr-release') {
+                            steps {
+                                dir('foreman-packaging') {
+                                    withCredentials([file(credentialsId: 'theforeman-bot-copr', variable: 'copr_config')]) {
+                                        obal(
```
evgeni, line 49: could/should this happen in parallel with the other (koji) release?

seems like a waste of time to wait on two external build systems in sequence.

ehelms: I think git-annex and building the SRPM might blow up since it's the same check out. I guess I could then make it fully parallel from copy-artifact through checkout and build.

evgeni: I always thought Jenkins magically create COW-copies of such folders when things run in parallel?
But fair,

evgeni (approved):

PR #369, opened 2023-11-01
Use upload_stage_rpms to rsync to stagingyum

ehelms:


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -479,7 +479,9 @@ def rsync_yum(user, ssh_key, collection, target, version) {
 
         sh """
             export RSYNC_RSH="ssh -i ${ssh_key}"
-            /usr/bin/rsync --archive --verbose --partial --one-file-system --delete-after ${collection}/${version} ${target_path}
+            export VERSION=${version}
+            export PROJECT=${collection}
+            ./upload_stage_rpms
```
ekohl, line 484: Don't you need to clone `theforeman-rel-eng` for the script?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -479,7 +479,9 @@ def rsync_yum(user, ssh_key, collection, target, version) {
```
ekohl, line 479: the line above this (`def target_path`) is now redundant. Same for the hosts definition. In fact, it can drop the whole loop.

ekohl (changes requested):

ehelms:
This should be good to go now that the nightly changes were merged to theforeman-rel-eng. I tested this (https://ci.theforeman.org/job/candlepin-nightly-rpm-pipeline/) for Candlepin pipeline. My intent will be to follow this change up with adopting a few other places to use the scripts rather than stand-alone code and centralize the logic into one place.

ehelms:
@ekohl could you take another look?

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -472,22 +472,17 @@ def rsync_debian(user, ssh_key, suite, component, deb_paths) {
...
-    for(host in hosts) {
-        def target_path = "${user}@${host}:rsync_cache/${target}/${version}/"
-
-        sh """
-            export RSYNC_RSH="ssh -i ${ssh_key}"
-            /usr/bin/rsync --checksum --times --perms --recursive --links --verbose --partial --one-file-system --delete-after ${collection}/${version}/ ${target_path}
-        """
+    if (!fileExists('upload_stage_rpms')) {
+        git url: "https://github.com/theforeman/theforeman-rel-eng", poll: false
     }
```
evgeni, line 480: I know, this is contrary to what Ewoud asked for, but the previous step is always `build_stage_repository` which comes from the same repo. So I'd argue it's either guaranteed that the repo is already checked out (one could also argue to have a prep step that does only that, for clarity), or whatever is on disk should absolutely not be synced.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -472,22 +472,17 @@ def rsync_debian(user, ssh_key, suite, component, deb_paths) {
...
+
+    sh """
+        export RSYNC_RSH="ssh -i ${ssh_key}"
```
ekohl, line 483: How is `ssh_key` defined?

ekohl (commented):
It took me a while to wrap my head around this, so I tried to see what I would do when I wrote this. I came up with https://github.com/theforeman/jenkins-jobs/pull/437. I think using `environment` instead of shell exports is a slightly cleaner way.

I was also tempted to write a function that creates the entire Jenkins stage with just 2 parameters.

In the short term I don't mind merging this if we can answer the `ssh_key` issue I noted inline. I'd also argue that eventually using the secrets mechanism in Jenkins (grep for `sshagent`) should be a better way because then we don't rely on Puppet providing the SSH key at some known location on some nodes. Not a blocker per se, but what I kept in mind writing the version I did.

theforeman.org/pipelines/lib/packaging.groovy
evgeni: line 476

ekohl: Bleh, reading diffs is hard.

evgeni: Yes. Blame Richmond.

PR #371, opened 2023-11-02
rsync foreman and katello to the correct production locations from st…

ehelms:
…agingyum

The Foreman and Katello repositories, when rsync from stagingyum need to have their destinations specified since it is not a 1-1 mapping between the two structures.

https://github.com/theforeman/foreman-infra/issues/1937 aims to address this longer term

### DIFF
--- a/theforeman.org/pipelines/lib/release.groovy
+++ b/theforeman.org/pipelines/lib/release.groovy
@@ -42,7 +42,17 @@ void mash(collection, version) {
 }
 
 void push_staging_rpms(repo_src, repo_dest, version, distro, keep_old_files = false) {
-    push_rpms_direct("${repo_src}/${distro}", "${repo_dest}/${version}/${distro}", !keep_old_files, keep_old_files, true)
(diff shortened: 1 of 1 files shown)


theforeman.org/pipelines/lib/release.groovy
```diff
@@ -42,7 +42,17 @@ void mash(collection, version) {
...
+        destination = "releases/${version}/${distro}"
+    } else if (repo_dest == 'katello') {
+        destination = "katello/${version}/katello/${distro}"
+    } else if (repo_dest == 'candlepin') {
+        destination = "katello/${version}/candlepin/${distro}"
+    } else {
+        destination = "${repo_dest}/${version}/${distro}"
+    }
+
+    push_rpms_direct("${repo_src}/${distro}", , !keep_old_files, keep_old_files, true)
```
evgeni, line 55: destination is unused?

ehelms: :eyes:

evgeni (approved):

PR #376, opened 2023-11-15
Refactor COPR jobs for easier branching

ekohl:
See individual commits for details. The goal is to use include files to create the right jobs so there is only a single one. For Foreman this is a bit uglier because it's more complex.


ehelms:
Right now the Copr PR test is locked to `rpm/develop` but would also need consideration for branching -- see https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/yaml/jobs/packaging/foreman-packaging-rpm-pr-test.yaml#L31C9-L31C30

ekohl:
> Right now the Copr PR test is locked to `rpm/develop` but would also need consideration for branching -- see https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/yaml/jobs/packaging/foreman-packaging-rpm-pr-test.yaml#L31C9-L31C30

Looking at the source, this is tricky in trying to reuse the other configs.

ehelms:
> Looking at the source, this is tricky in trying to reuse the other configs.

Yes... I think I would have to combine the jobs into a single job with `when` conditions. Question is if you want to block on that, or go the quick route and I can follow up with the re-factor.

ekohl:
I'm tempted to go the easy route on the triggers.

ehelms (approved):
If you want to address the PR test issue here or in another PR.

branch-foreman
```diff
@@ -21,6 +21,8 @@ sed -i "/nightly/i \ \ \ \ \ \ - '${KATELLO_VERSION}'" centos.org/jobs/katello-p
 
 echo "- '${FOREMAN_VERSION}'" >> theforeman.org/yaml/includes/foreman_versions.yaml.inc
```
evgeni, line 22: if we're not doing koji releases anymore, can't this go?

(also, the file is now foreman_versions_koji, right?)

ekohl: No, this is the file with **all** releases. It's also used for Debian and other stuff like `smart-proxy-X.Y-stable-test` & `test_X_Y_stable`.

ekohl:
I think this covers it by introducing new GHPRB triggers with regexes. Not pretty, but I'd expect it to work. On a related note: we should really get away from GHPRB since it's unmaintained and has security problems.

theforeman.org/yaml/includes/pulpcore_versions_copr.yaml.inc
```diff
@@ -0,0 +1 @@
+- 'nightly'
```
Odilhao, line 1: We can include '3.39' here as well, since we are going to branch from `rpm/develop`

ehelms (approved):

theforeman.org/yaml/includes/pulpcore_versions_copr.yaml.inc
ekohl: I think that's out of scope for this PR: we should treat 3.39 as a separate branching step. We already know https://github.com/theforeman/theforeman-rel-eng/blob/master/procedures/pulpcore/update.md.erb is out of date, so please keep notes when you do.

PR #381, opened 2023-11-20
rsync to stagingyum per distro

ehelms:
The rsync script on web01 needs the version to exist on the rsync target in order to properly know what release to work with. In order to include the release we need to rsync per OS rather than one command for the whole version.

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -475,11 +475,11 @@ def rsync_yum(user, ssh_key, collection, target, version) {
     def hosts = ["web01.osuosl.theforeman.org"]
 
     for(host in hosts) {
-        def target_path = "${user}@${host}:rsync_cache/${target}"
+        def target_path = "${user}@${host}:rsync_cache/${target}/${version}/"
 
         sh """
             export RSYNC_RSH="ssh -i ${ssh_key}"
-            /usr/bin/rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after ${collection}/${version} ${target_path}
+            /usr/bin/rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after ${collection}/${version}/ ${target_path}
         """
     }
 }


ekohl (changes requested):
The idea behind sync a whole version is that it's atomic. For Debian (where we have multiple directories) we join the paths:
https://github.com/theforeman/jenkins-jobs/blob/d78f0f3b828444b109105fa5d27927be3f158c31/theforeman.org/pipelines/lib/packaging.groovy#L462

That way it's a single atomic action and everything gets promoted as one. You don't have the risk (*) that EL9 is newer than EL8 because the network failed after EL9 was done.

I still think that we should solve it similar to https://github.com/theforeman/foreman-infra/pull/1994 instead. I'll leave some comments there.

* Technically, everything can crash, but you minimize the risk

evgeni:
> The idea behind sync a whole version is that it's atomic. For Debian (where we have multiple directories) we join the paths:
> 
> https://github.com/theforeman/jenkins-jobs/blob/d78f0f3b828444b109105fa5d27927be3f158c31/theforeman.org/pipelines/lib/packaging.groovy#L462

While the above excerpt looks like it, we don't do that. `deb_paths` are the paths of the actual debs.

The place where we actually call rsync methods from is here:
https://github.com/theforeman/jenkins-jobs/blob/d78f0f3b828444b109105fa5d27927be3f158c31/theforeman.org/pipelines/lib/packaging.groovy#L186-L197

And `repos` has keys of the form `${suite}-${component}`, which is `bullseye-3.9`, `focal-plugins` etc.

You can also see it in Jenkins where we have separate "push-{version}-{os}" steps for stage→prod:
https://ci.theforeman.org/blue/organizations/jenkins/foreman-nightly-deb-pipeline/detail/foreman-nightly-deb-pipeline/1353/pipeline
https://ci.theforeman.org/blue/organizations/jenkins/foreman-3.8-release-pipeline/detail/foreman-3.8-release-pipeline/6/pipeline/166

And "{os}-{version}" for packaging → stage:
https://ci.theforeman.org/blue/organizations/jenkins/foreman-packaging-deb-3.8-release/detail/foreman-packaging-deb-3.8-release/12/pipeline/131

ehelms:
So what's the verdict on this PR? Pipelines continue to be broken without a
resolution.

theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -475,11 +475,11 @@ def rsync_yum(user, ssh_key, collection, target, version) {
     def hosts = ["web01.osuosl.theforeman.org"]
 
     for(host in hosts) {
-        def target_path = "${user}@${host}:rsync_cache/${target}"
+        def target_path = "${user}@${host}:rsync_cache/${target}/${version}"
 
         sh """
             export RSYNC_RSH="ssh -i ${ssh_key}"
-            /usr/bin/rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after ${collection}/${version} ${target_path}
+            /usr/bin/rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after --mkpath tmp/${collection}/${version}/* ${target_path}
```
ekohl, line 482:
```suggestion
            /usr/bin/rsync --checksum --perms --recursive --links --verbose --partial --one-file-system --delete-after --mkpath ${collection}/${version}/ ${target_path}/
```

ekohl (changes requested):
I was playing a bit with it and didn't finish it yesterday, but in https://github.com/theforeman/foreman-infra/pull/1994#issuecomment-1822738006 I wrote down my findings.

The short version is to add trailing slashes, so my recommendation is to `rsync ${collection}/${version}/ ${user}@${host}:rsync_cache/${target}/${version}/`

ekohl (approved):
I didn't test this out end-to-end, but since it's what I recommended based on local testing I approve :)

PR #382, opened 2023-11-24
Use artifacts from the upstream job, if possible

ekohl:
We've seen multiple times that a package release job runs, but doesn't use the latest source. My theory is that there's a race condition between declaring the last source release job as last successful and the package release job running.

The Copy Artifact Plugin has an [option for the selector](https://www.jenkins.io/doc/pipeline/steps/copyartifact/). That can be set to upstream with a fallback to the last successful. That would mean that when a source release job triggers a package release job, the triggering job is the upstream. If there is none (like with a manual run), then it should fall back to the last successful.

I'm unsure about the syntax, but https://stackoverflow.com/questions/53820503/how-to-copy-artifacts-from-other-jenkins-job-from-a-pipeline/53821560#53821560 suggests the `selector: upstream()` part is correct and I'm assuming the option is passed this way.

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -259,7 +259,7 @@ def setup_sources_core(project, os, version, repoowner, pull_request = false) {
         def build_vars = read_build_vars(project)
         if (version == 'nightly') {
             if (build_vars['source_location']) {
-                copyArtifacts(projectName: build_vars['source_location'], excludes: 'package-lock.json', flatten: true)
+                copyArtifacts(projectName: build_vars['source_location'], excludes: 'package-lock.json', flatten: true, selector: upstream(fallbackToLastSuccessful: true))
                 sh(script: "mv *.tar.bz2 ${project}_${package_version}.orig.tar.bz2", label: "rename tarball")
                 last_commit = readFile('commit').trim()
             } else {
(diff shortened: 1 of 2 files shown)


ekohl:
> My theory is that there's a race condition between declaring the last source release job as last successful and the package release job running.

Now that I investigate it, this isn't right.

https://ci.theforeman.org/job/foreman-installer-develop-package-release/1576/ was triggered by https://ci.theforeman.org/job/foreman-installer-develop-source-release/1657/ and it also states:

> [2023-11-24T03:11:48.009Z] Copied 3 artifacts from "foreman-installer-develop-source-release" build number 1657

And if I look in the tarball that 1657 uses, then it has the correct changes included (I checked `README.md`).

Yet https://github.com/theforeman/puppet-pulpcore/pull/315#issuecomment-1826084380 proves nightly didn't.

Also, the resulting SRPM from https://ci.theforeman.org/job/foreman-installer-develop-package-release/1576/ has the correct changes (again, checked `README.md`).

So `foreman-installer-3.9.0-0.3.develop.20231124031230gite0bf154.src.rpm` should be good. Looking at https://stagingyum.theforeman.org/foreman/nightly/el8/x86_64/, we see https://stagingyum.theforeman.org/foreman/nightly/el8/x86_64/foreman-installer-3.9.0-0.3.develop.20231124031230gite0bf154.el8.noarch.rpm there and that also contains the correct `README.md`.

We can also see that https://ci.theforeman.org/job/foreman-nightly-rpm-pipeline/2098/ synced it and included the change (and passed).

Yet on https://yum.theforeman.org/nightly/el8/x86_64/ I see https://yum.theforeman.org/nightly/el8/x86_64/foreman-installer-3.9.0-0.3.develop.20231123031245gite0bf154.el8.noarch.rpm so somehow it's not being synced from staging to release.

PR #383, opened 2023-11-26
Drop Pulpcore 3.16, 3.17 & 3.18 pipelines

ekohl:
These belong to EOL Katello versions. Recently we've been seeing problems provisioning nodes in CentOS CI so this reduces the number of nodes we consume.


evgeni (approved):
I think you can also drop 3.21

ekohl:
I wondered if there was a downstream concern, since it's used in Foreman 3.5

evgeni:
I'll leave the definitive answer to @Odilhao but I don't think the upstream pipes matter for that case

Odilhao:
Let's nuke it and keep 3.22, 3.28 and 3.39, too many python 3.9 to cherrypick and release atm.

evgeni (approved):

PR #385, opened 2023-11-29
fix 3.9 jenkins job definitions

evgeni:
Fixes: 55c177cca3c740e7dae231b3fbcb7f51025f1e53


theforeman.org/pipelines/vars/foreman/3.9.groovy
```diff
@@ -1,12 +1,11 @@
 def foreman_version = '3.9'
 def git_branch = "${foreman_version}-stable"
-def stage_source = 'copr'
+def stage_source = 'stagingyum'
 
 def foreman_client_distros = [
     'el9',
     'el8',
-    'el7',
-    'sles12'
```
ekohl, line 9: I didn't realize we'd effectively drop sles12 this way. I wonder if we could somehow restore this at some point, like syncing from OBS.

evgeni: copr can build for sles, so we could… just build it, properly.

theforeman.org/yaml/includes/foreman_versions_copr.yaml.inc
```diff
@@ -1 +1 @@
-[]
+- '3.9'
```
ekohl, line 1: I'd have expected this line to add this:
https://github.com/theforeman/jenkins-jobs/blob/b993393b01c86082cefda9cfea5dfbba5abd0d3c/branch-foreman#L24

theforeman.org/yaml/jobs/pipeline/katello-rpm-pipeline.yaml
```diff
@@ -20,3 +20,4 @@
     version:
       - '4.9'
       - '4.10'
+      - '4.11'
```
ekohl, line 23: This should be covered:
https://github.com/theforeman/jenkins-jobs/blob/b993393b01c86082cefda9cfea5dfbba5abd0d3c/branch-foreman#L26

theforeman.org/yaml/views/release.yml
```diff
@@ -13,6 +13,7 @@
     minor:
         - '7'
         - '8'
+        - '9'
```
ekohl, line 16: This is called out as a manual TODO:
https://github.com/theforeman/jenkins-jobs/blob/b993393b01c86082cefda9cfea5dfbba5abd0d3c/branch-foreman#L31-L32

theforeman.org/pipelines/vars/foreman/3.9.groovy
ekohl: It's not listed in nightly, which is the source for this:
https://github.com/theforeman/jenkins-jobs/blob/b993393b01c86082cefda9cfea5dfbba5abd0d3c/theforeman.org/pipelines/vars/foreman/nightly.groovy#L5-L9

ekohl (commented):
I did a more careful analysis of where it went wrong, but I can't explain it. It should all be covered. The PR was also made on top of a commit that had the correct script. The only thing I can think of is that it was made against an older version and then rebased before submitted as a PR.

evgeni:
yeah, I have no idea how this happened.

PR #390, opened 2023-12-15
properly set dist when inspecting copr info

evgeni:
otherwise repoclosure runs with "dist: null"


evgeni:
see https://github.com/theforeman/jenkins-jobs/blob/9ce7bbc07364cd0cfcd13df0de6a63c762d63647/theforeman.org/pipelines/test/rpm_copr_packaging.groovy#L168

ehelms (approved):

PR #391, opened 2023-12-15
For Ruby 2.7 tests use dedicated bundler version

ehelms:
Bundler 2.4.22 is last version to support Ruby 2.7 and needs to be directly installed to prevent an error resolving the supported version. This requirement occurred with the release of bundler 2.5.0.

### DIFF
--- a/theforeman.org/pipelines/lib/rvm.groovy
+++ b/theforeman.org/pipelines/lib/rvm.groovy
@@ -15,6 +15,11 @@ def gemset(name = null) {
 
 def configureRVM(ruby, name = '', bundler_version = null) {
     emptyGemset(ruby, name)
+
+    if (ruby == '2.7') {
+        bundler_version = '2.4.22'
+    }
+
     if (bundler_version) {
         withRVM(["gem install bundler -v '${bundler_version}' --no-document"], ruby, name)
     } else {


theforeman.org/pipelines/lib/rvm.groovy
```diff
@@ -15,6 +15,11 @@ def gemset(name = null) {
 
 def configureRVM(ruby, name = '', bundler_version = null) {
     emptyGemset(ruby, name)
+
+    if (ruby == '2.7') {
+        bundler_version = '2.4.22'
```
evgeni, line 20: Should this be `< 2.5` to allow 2.4.z patches?

ehelms: No idea. But since you are asking, I am assuming maybe yes.

evgeni: :)

I have no idea how high the chances of a 2.4.z is, but playing safe is always good :)

ekohl:
non-pipeline jobs are still failing. https://github.com/theforeman/jenkins-jobs/tree/master/theforeman.org/scripts/test needs to be updated. @pcreech is working on that now.

PR #394, opened 2024-01-03
Use bundler 2.4.22 in ruby 2.7 environments

pcreech:
This extends the solution for commit
1d4dd26f8e27920534b806321b21fd31c16ee96b to the shell script based jobs

### DIFF
--- a/theforeman.org/yaml/builders/smart-proxy-plugin.yaml
+++ b/theforeman.org/yaml/builders/smart-proxy-plugin.yaml
@@ -13,7 +13,12 @@
           rvm gemset empty --force
           set -x
 
-          gem install bundler --no-document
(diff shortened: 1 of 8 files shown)


theforeman.org/yaml/builders/smart-proxy-plugin.yaml
```diff
@@ -13,7 +13,12 @@
           rvm gemset empty --force
           set -x
 
-          gem install bundler --no-document
+          if ["${{ruby}}" = '2.7']
```
pcreech, line 16: Annoyingly, with the quoted string in bash, we needed to manually do the double bracketing instead of jenkins job builder doing it for us in the .yaml configs.  This was the cause of previous job failures.

ekohl (approved):

PR #395, opened 2024-01-04
enable nightly el9 repo generation, closure and publishing

evgeni:
this will *NOT* perform any tests on el9 yet


evgeni:
shouldn't be merged before https://github.com/theforeman/foreman-packaging/pull/10200, as otherwise plugin pipes will start failing (katello has an own definition of "which EL versions is Foreman available on", but plugins rely on the list here, and thus would start doing the same steps

ehelms (approved):

PR #397, opened 2024-01-04
Drop kafo tests

ehelms:
Kafo testing is handled by Github Actions.


ekohl (approved):
We should also make sure we can up any webhooks, if any.

And set up GHA as a required action. To do so, I've opened https://github.com/theforeman/kafo/pull/370.

PR #399, opened 2024-01-05
Fix ruby version check in plugin test

adamruzicka:
Before
```
$ if ["${ruby}" = '2.7']; then echo yay; fi
bash: [2.7: command not found
```

After
```
$ if [ "${ruby}" = '2.7' ]; then echo yay; fi
yay
```

Fixes: f164a3f291045dd45fac4dd560421f243af1fbc3


adamruzicka:
Hopefully should address this kind of failure https://ci.theforeman.org/job/foreman_openscap-pull-request/531/label=fast,ruby=2.7/console

ekohl (approved):
I was debating this when I reviewed it. @pcreech guess I should have commented after all ;)

PR #400, opened 2024-01-05
Record RuboCop issues for Katello

ekohl:
This changes the invocation to directly call RuboCop instead of relying on a rake task. This is copied from smart-proxy.groovy.

Interstingly enough, it doesn't use the katello:rubocop:jenkins task that also exists.

I'm not 100% sure about the env.WORKSPACE part, but we need to point it to the actual checkout and I believe that's the right variable.


ehelms (approved):

evgeni:
we use GHA now, closing

PR #401, opened 2024-01-05
Use correct braces in smart-proxy-plugin script

ekohl:
These are variables that are replaced in Jenkins templating, not bash.

Fixes: f164a3f291045dd45fac4dd560421f243af1fbc3 ("Use bundler 2.4.22 in ruby 2.7 environments")

### DIFF
--- a/theforeman.org/yaml/builders/smart-proxy-plugin.yaml
+++ b/theforeman.org/yaml/builders/smart-proxy-plugin.yaml
@@ -13,7 +13,7 @@
           rvm gemset empty --force
           set -x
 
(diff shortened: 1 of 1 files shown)


evgeni (approved):

evgeni:
> KeyError: 'ruby'

huh…

ekohl:
Layers of templating ...

evgeni:
Shall we just always instal that bundler version? It works fine with ruby3 too…

ekohl:
I'm unsure about the Foreman plugin testing, but this looks correct for Smart Proxy. For Foreman plugins I'd rather invest time in phasing it out.

PR #402, opened 2024-01-08
Fix substitution syntax

adamruzicka:
To fix errors like
```
/tmp/jenkins415375971260303110.sh: line 12: ${{ruby}}: bad substitution
```

seen in https://ci.theforeman.org/job/smart-proxy-openscap-pull-request/60/ruby=2.7/console


adamruzicka:
Oh, looks like I'm just redoing a subset of #401 . Nevermind then

PR #403, opened 2024-01-11
add el9 to katello repository generation/publishing

evgeni:


evgeni:
draft until https://github.com/theforeman/foreman-packaging/pull/10269 (or a similar solution) is merged, as otherwise repoclosure will prevent *any* publishing

ekohl (approved):

PR #405, opened 2024-01-12
add bookworm to foreman nightly

evgeni:


evgeni:
draft, as this will start trying to build nightly packages on those releases too, but we're not there yet.

theforeman.org/pipelines/vars/foreman/nightly.groovy
```diff
@@ -11,11 +11,12 @@ def foreman_el_releases = [
     'el9',
     'el8'
 ]
-def foreman_debian_releases = ['bullseye', 'focal', 'jammy']
+def foreman_debian_releases = ['bullseye', 'bookworm', 'focal', 'jammy']
 
 def pipelines_deb = [
     'install': [
         'debian11',
+        'debian12',
```
evgeni, line 19: The pipelines currently fail because there is no `puppetserver` for bookworm on apt.puppet.com and the one in Debian does not install when using `puppet-agent` from apt.puppet.com, and I really didn't want to untangle that AIO mess today.

evgeni: I propose we add bookworm to `debian_releases`, which will ensure it's built and published, and then see when we can actually test it.

evgeni: https://github.com/theforeman/jenkins-jobs/pull/433

ekohl (approved):

PR #406, opened 2024-01-12
Add EL9 to Candlepin nightly

ehelms:


theforeman.org/pipelines/vars/candlepin/nightly.groovy
```diff
@@ -1,8 +1,12 @@
...
 def packaging_branch = 'rpm/develop'
+def candlepin_distros = [
+    'el8',
+    'el9'
+]
 def pipelines = [
     'candlepin': [
-        'centos8-stream'
+        'centos8-stream',
+        'centos9-stream'
```
evgeni, line 10: This is currently unused (as we do not yet have CP tests), but would start failing once we do, or does this PR mean we also have el9 packages now, not only repos?

evgeni (approved):

theforeman.org/pipelines/vars/candlepin/nightly.groovy
evgeni: https://copr.fedorainfracloud.org/coprs/g/theforeman/candlepin-nightly-staging/build/6888982/ seems like there are EL9 builds 🤘

ehelms: Means both -- I tested this before hand with nightly https://yum.theforeman.org/candlepin/nightly/el9/x86_64/

ehelms: And working puppet module tests -- https://github.com/theforeman/puppet-candlepin/pull/251

PR #407, opened 2024-01-16
Add Candlepin pipeline tests

ehelms:


evgeni (approved):

evgeni:
well, ack once jjb can actually generate this, but ack on the approach

PR #408, opened 2024-01-16
Set Candlepin pipeline name correctly

ehelms:


ekohl (approved):

theforeman.org/pipelines/release/pipelines/candlepin.groovy
```diff
@@ -52,7 +52,7 @@ pipeline {
 
             steps {
                 script {
-                    runDuffyPipeline('candlepin', candlepin_version)
+                    runDuffyPipeline('candlepin-rpm', candlepin_version)
```
ekohl, line 55: Actually, I think this refers to a line which I think contains a problem: https://github.com/theforeman/jenkins-jobs/commit/952742de34dcb06a3464df53c0f2f676fd179d1e#r137179772

ekohl (changes requested):

theforeman.org/pipelines/release/pipelines/candlepin.groovy
ekohl: Going through what I checked: `runDuffyPipeline` will expand the job name to `foreman-pipeline-candlepin-rpm-$version` which matches.

ekohl (approved):

PR #416, opened 2024-02-16
Always cleanup gemsets

ehelms:


theforeman.org/scripts/gemset_cleanup.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-[ -z "$ruby" ] && ruby=2.0.0
+[ -z "$ruby" ]
```
ekohl, line 2: This was a confusing statement, but it basically came down to:
```suggestion
if [ -z "$ruby" ] ; then
  ruby=2.0.0
fi
```

So I'm not sure how this is supposed to improve things. You'll just end up with `rvm use ruby-@${gemset}` which is incorrect.

ehelms: Ha, yea that was not obvious. Cleaned up and added an actual cleanup.

evgeni:
mostly not needed anymore, and last use of cleanup_gemset is removed in https://github.com/theforeman/jenkins-jobs/pull/452

PR #419, opened 2024-02-21
Use python3.11 on EL8 for package json filtering

ehelms:


ianballou:
Just popping in to say this is required for us to release Katello 4.11.1

evgeni:
Why don't we run this with whatever python3 is available?

ehelms:
> Why don't we run this with whatever python3 is available?

*shrug* I just copied what we do here -- https://github.com/theforeman/jenkins-jobs/blob/8a5a5f9e46a7a807356c8c87c6033aa2ff43a323/theforeman.org/pipelines/lib/obal.groovy#L10

evgeni:
Yeah, but that needs ansible-python, not any python ;)

PR #425, opened 2024-02-28
only trigger koji rpm builds for rpm/3.8

evgeni:
`[0-8]+` also matches `10` ;)

### DIFF
--- a/theforeman.org/yaml/triggers/github_pr.yaml
+++ b/theforeman.org/yaml/triggers/github_pr.yaml
@@ -39,7 +39,7 @@
           <<: *ghprb_defaults
           status-url: https://ci.theforeman.org/blue/organizations/jenkins/$JOB_NAME/detail/$JOB_NAME/$BUILD_ID/pipeline
           white-list-target-branches:
-            - rpm/3\.[0-8]+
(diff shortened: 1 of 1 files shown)


ehelms (approved):

theforeman.org/yaml/triggers/github_pr.yaml
```diff
@@ -39,7 +39,7 @@
           <<: *ghprb_defaults
           status-url: https://ci.theforeman.org/blue/organizations/jenkins/$JOB_NAME/detail/$JOB_NAME/$BUILD_ID/pipeline
           white-list-target-branches:
-            - rpm/3\.[0-8]+
+            - rpm/3\.8
```
ekohl, line 42: Wouldn't a more correct fix have been to only drop `+`? I'm fine with it now, until we need to revive 3.7 ;)

evgeni: Let's just not revive it then? ;)

ekohl: I would have also accepted "He's dead, Jim"

PR #431, opened 2024-03-14
Drop Foreman plugin tests

ekohl:
Both foreman_hooks and foreman_virt_who_configure have been converted.

foreman_default_hostgroup is being worked on, but has been broken for 3 years so nothing is lost there.

foreman_setup is being retired and can already be dropped.


ekohl:
https://ci.theforeman.org/view/Plugins/ can be dropped when this is merged.

evgeni (approved):

evgeni:
Katello uses own test definitions, in Groovy, we can drop those once we don't support any of the stable branches.

ekohl:
> we can drop those once we don't support any of the stable branches.

Or cherry pick the GHA workflows?

PR #437, opened 2024-03-25
Rely on theforeman-rel-eng projects to upload staging

ekohl:
This introduces simple wrappers around theforeman-rel-eng scripts to do the heavy lifting. The biggest benefit is that more configuration is removed from this repository and pushed into theforeman-rel-eng where it was already anyway.

It is an alternative to https://github.com/theforeman/jenkins-jobs/pull/369. The biggest difference is using environment variables via native Jenkins to separate out the logic and keep method invocations easier. Right now it collapses two stages into a single stage since they both share the same configuration.

I'll reference this in my review of https://github.com/theforeman/jenkins-jobs/pull/369.

### DIFF
--- a/theforeman.org/pipelines/lib/packaging.groovy
+++ b/theforeman.org/pipelines/lib/packaging.groovy
@@ -481,22 +481,19 @@ def rsync_debian(user, ssh_key, suite, component, deb_paths) {
     }
 }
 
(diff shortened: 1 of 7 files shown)


evgeni:
Do you still want to look into this, given #369 is merged?

PR #442, opened 2024-04-11
Test using rbenv for foreman-installer tests

ehelms:
There are a couple implications here:

 1. Using a node wide cache for rubygems to speed bundle install up
 2. Using shell based rbenv to prevent clashes between instances

### DIFF
--- a/theforeman.org/pipelines/lib/rbenv.groovy
+++ b/theforeman.org/pipelines/lib/rbenv.groovy
@@ -1,0 +1,30 @@
+def bundleInstall(version, gemfile=null) {
+    command = "bundle install"
+
+    if (gemfile) {
+        command = "${command} --gemfile=${gemfile}"
+    }
+
+    withRuby(version, "bundle config set path ~/.rubygems")
+    withRuby(version, command)
+}
+
+def bundleExec(version, command, gemfile=null) {
+    command = "bundle exec ${command}"
+
+    if (gemfile) {
+        command = "BUNDLE_GEMFILE=${gemfile} ${command}"
+    }
+
+    withRuby(version, command)
+}
+
+def withRuby(version, command) {
+    echo command.toString()
+
+    sh """
+        export PATH="\$HOME/.rbenv/shims:\$PATH"
+        export RBENV_VERSION=${version}
+        ${command}
+    """
+}
--- a/theforeman.org/pipelines/test/foreman-installer.groovy
+++ b/theforeman.org/pipelines/test/foreman-installer.groovy
@@ -12,7 +12,7 @@ pipeline {
                 axes {
                     axis {
                         name 'ruby'
-                        values '2.7'
+                        values '2.7.6'
                     }
                     axis {
                         name 'PUPPET_VERSION'
@@ -23,7 +23,7 @@ pipeline {
                     exclude {
                         axis {
                             name 'ruby'
-                            notValues '2.7'
+                            notValues '2.7.6'
                         }
                         axis {
                             name 'PUPPET_VERSION'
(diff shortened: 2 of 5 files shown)


theforeman.org/pipelines/lib/rbenv.groovy
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    withRuby(version, command)
+}
+
+def withRuby(version, command) {
+    echo command.toString()
+
+    sh """
+        eval "\$(rbenv init -)"
```
ehelms, line 26: This does not seem to work to do this per shell rather than doing it for the whole of the node.

ekohl: My theory is that it has to do with interactive vs non-interactive shell modes. https://github.com/rbenv/rbenv/wiki/Unix-shell-initialization has a nice overview, but doesn't explain if you need to pass some options for non-interactive mode.

theforeman.org/pipelines/lib/rbenv.groovy
```diff
@@ -0,0 +1,31 @@
...
+    command = "bundle exec ${command}"
+
+    if (gemfile) {
+        command = "BUNDLE_GEMFILE=${gemfile} bundle exec ${command}"
+    }
+
+    withRuby(version, command)
+}
+
+def withRuby(version, command) {
```
ehelms, line 22: An alternative approach would be to include a check for if a particular environment variable is set (e.g. RUBY_VERSION) then use it unless version is explicitly specified. That would allow not having to pass the ruby version as the first argument everytime if we find that annoying.

ehelms:
This initial change includes:

- Set of shared library to setup and run in rbenv context
- conversion of foreman-installer-pr-test job to show how it will look
- a simple ruby test job that allows more easily testing changes to how Ruby test code works

theforeman.org/pipelines/lib/rbenv.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+    withRuby(version, "bundle config set path ~/.rubygems")
+    withRuby(version, command)
+}
+
+def bundleExec(version, command, gemfile=null) {
+    command = "bundle exec ${command}"
+
+    if (gemfile) {
+        command = "BUNDLE_GEMFILE=${gemfile} bundle exec ${command}"
```
evgeni, line 16: this leads to the command being `BUNDLE_GEMFILE=lol bundle exec bundle exec command`

theforeman.org/pipelines/lib/rbenv.groovy
```diff
@@ -0,0 +1,31 @@
...
+    withRuby(version, command)
+}
+
+def withRuby(version, command) {
+    echo command.toString()
+
+    sh """
+        export PATH="/usr/lib64/rbenv/libexec/:\$HOME/.rbenv/shims:\$HOME/.rbenv/versions:\$PATH"
+        export RBENV_VERSION=${version}
+        export RBENV_ROOT="\$HOME/.rbenv"
```
evgeni, line 28: I would have expected the following is sufficient, but did not further test it:
```suggestion
        export PATH="\$HOME/.rbenv/shims:\$PATH"
        export RBENV_VERSION=${version}
```

theforeman.org/pipelines/lib/rbenv.groovy
ehelms: Double your bundle for safety.

theforeman.org/pipelines/lib/rbenv.groovy
ehelms: You are correct -- https://ci.theforeman.org/job/simple-ruby-test/28/console

evgeni (approved):

PR #445, opened 2024-04-21
Convert jobs to rbenv

ehelms:


theforeman.org/pipelines/lib/rbenv.groovy
```diff
@@ -1,8 +1,8 @@
-def bundleInstall(version, gemfile=null) {
-    command = "bundle install"
+def bundleInstall(version, options=null) {
+    command = "bundle install --jobs 5 --retry 5"
```
evgeni, line 2: `theforeman.org/pipelines/test/foreman-installer.groovy` and `theforeman.org/pipelines/test/simple-ruby.groovy` are using the `gemfile` param and would break with this change.

evgeni (changes requested):
LGTM besides the breakage to the existing two jobs already using rbenv

theforeman.org/pipelines/lib/rbenv.groovy
ehelms: Ahh yea, thanks. Got it.

evgeni (approved):

PR #451, opened 2024-05-08
drop hammer_cli(_foreman) PR testing

evgeni:
this is now done on GitHub Actions


ekohl (approved):
I think we can rely on GHA.

PR #454, opened 2024-05-08
use installation tests rake command instead of hard-coding

evgeni:


ekohl (approved):
I like this, but we should cherry pick https://github.com/theforeman/foreman-installer/commit/a500834d5a154303ed671396c2e4a2d530178bad to our stable branches.

Also, we should really add Puppet 8 to the mix.

PR #457, opened 2024-05-08
drop foreman-installer-pr-test

evgeni:
migrated to GHA

needs https://github.com/theforeman/foreman-installer/pull/944 and https://github.com/theforeman/foreman-installer/pull/943


evgeni:
I've even fixed the branch protection rules! :tada:

PR #458, opened 2024-05-08
Align RPM pipelines' groovy code

ekohl:
This attempts to align the code as much as possible. See individual commits for details.

My goal would be to create some common library function that generates the whole pipeline with some parameters so we don't need to copy things.

Big differences remaining:
* Repoclosures rely on variables. Perhaps we can also utilize theforeman-rel-eng here (since it already knows how to create and sync the repos).
* Production repo pushing. Again, theforeman-rel-eng is a good candidate to use.

Currently very much untested and the SSH key needs some manual work to be created with some clean up in foreman-infra as well.


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -481,17 +481,12 @@ def rsync_debian(user, ssh_key, suite, component, deb_paths) {
...
-    }
+    sh "./generate_stage_repository"
 
-    sh """
-        export RSYNC_RSH="ssh -i ${ssh_key}"
-        export VERSION=${version}
-        export PROJECT=${collection}
-        ./upload_stage_rpms
-    """
+    sshagent(['yum-stage']) {
```
ekohl, line 489: This will need some manual migration, but allows us to drop this bit:
https://github.com/theforeman/foreman-infra/blob/801376af314f44108b4a851edc9e2d2173f7c247/puppet/modules/jenkins_node/manifests/packaging/rpm.pp#L44-L48

theforeman.org/pipelines/release/pipelines/candlepin.groovy
```diff
@@ -8,35 +8,29 @@ pipeline {
...
-                    candlepin_distros.each { distro ->
-                        sh "./build_stage_repository candlepin ${candlepin_version} ${distro}"
-                    }
-                }
-            }
-        }
-        stage('staging-copy-repository') {
+    script {
+        env.VERSION = candlepin_version
+    }
```
ekohl, line 17: I can't find a good way to set the environment variable as a script. Perhaps this needs to be
```suggestion
    env.VERSION = candlepin_version
```

theforeman.org/pipelines/release/pipelines/plugins.groovy
```diff
@@ -8,29 +8,31 @@ pipeline {
...
+        env.DISTROS = foreman_el_releases
+        env.VERSION = foreman_version
+    }
+
     stages {
-        stage('staging-build-repository') {
-            steps {
-                git url: "https://github.com/theforeman/theforeman-rel-eng", poll: false
+        stage('staging-repository') {
+            // No when condition because it's not signed
```
ekohl, line 24: This is the odd one out here

ekohl (commented):
I've tried to keep the diff between all pipelines to just the `environment` and `script` blocks at the top (and the timeout). Should we take it a step further and create a function to build an RPM pipeline with a few parameters?

PR #460, opened 2024-05-10
drop deploy_web

evgeni:


evgeni:
this might be a bit early

https://github.com/theforeman/foreman-infra/issues/1724#issuecomment-2104190465

PR #461, opened 2024-05-10
use python 3.11 for jjb

evgeni:


ekohl (commented):
Do we have 3.11 installed on our Jenkins nodes?

evgeni:
Yes, Ansible/obal use that

ekohl (approved):

PR #464, opened 2024-05-20
Foreman 3.11 Jobs/Pipelines

zjhuntin:


theforeman.org/yaml/views/release.yml
```diff
@@ -15,6 +15,7 @@
         - '8'
```
ekohl, line 15: Unrelated, but we should clean up 7 & 8 here. Looks like they don't show up in the UI, so it doesn't matter now, but still nice to clean up.

ekohl (approved):

PR #468, opened 2024-05-22
Remove use of gemset variable

zjhuntin:


evgeni (approved):

evgeni (changes requested):
wha, me--

but this needs to be "unique"

zjhuntin:
Ahh, my bad

theforeman.org/scripts/test/test_develop.sh
```diff
@@ -32,11 +32,11 @@ bundle exec rake rubocop
 
 # Database environment
 (
```
evgeni, line 34: can you replace `gemset` with something like `jobid` and do before this block
```
jobid=$(echo ${JOB_NAME} | cut -d/ -f1)-${EXECUTOR_NUMBER}
```

the databases need to be unique

evgeni:
> Ahh, my bad

nope, that's on me via 93ce4d5b01b11fe8e89151e4342e0cbbdd459f51

evgeni (approved):

PR #469, opened 2024-05-23
make the chroot_to_dist method more generic

evgeni:
this means we do not have to add entries for new versions, just for new
distributions or families

it also drops the `return null` fall-through, triggering an error when
an unknown chroot is passed


theforeman.org/pipelines/lib/copr.groovy
```diff
@@ -40,15 +40,11 @@ def copr_repos(package_name) {
...
-    } else if(chroot == 'opensuse-leap-15.5-x86_64') {
-       return 'leap155'
-    } else {
-        return null
-    }
+    def family_map = ['rhel': 'el', 'centos-stream': 'el', 'opensuse-leap': 'leap']
+
+    chroot_parts = chroot.split('-')
+    version = chroot_parts[-2].replace('.', '')
+    os = chroot_parts[0..-3].join('-')
```
ekohl, line 47: My brain prefers this the other way around so you first take the beginning of the string and then the end.
```suggestion
    os = chroot_parts[0..-3].join('-')
    version = chroot_parts[-2].replace('.', '')
```

ekohl (approved):

theforeman.org/pipelines/lib/copr.groovy
evgeni: your brain is weird :P

ekohl: Tell me something I don't know ;)

PR #474, opened 2024-06-03
Drop centos8-stream support from pipelines

archanaserver:
Updated pipeline configuration in [theforeman.org/pipelines/vars/katello/](https://github.com/theforeman/jenkins-jobs/tree/master/theforeman.org/pipelines/vars/katello) and [theforeman.org/pipelines/vars/foreman/](https://github.com/theforeman/jenkins-jobs/tree/master/theforeman.org/pipelines/vars/foreman) to drop support for `centos8-stream` due to CentOS 8 EOL

Dropped centos8-stream support and added AlmaLinux support for candlepin in theforeman.org/pipelines/vars/candlepin and pulpcore in theforeman.org/pipelines/vars/pulpcore.

### DIFF
--- a/theforeman.org/pipelines/vars/candlepin/4.3.groovy
+++ b/theforeman.org/pipelines/vars/candlepin/4.3.groovy
@@ -3,7 +3,8 @@ def candlepin_distros = ['el8', 'el9']
 def packaging_branch = 'rpm/4.3'
 def pipelines = [
     'candlepin': [
-        'centos8-stream',
-        'centos9-stream'
+        'centos9-stream',
+        'almalinux8',
+        'almalinux9',
     ]
 ]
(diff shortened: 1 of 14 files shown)


ekohl (changes requested):
In itself it looks ok for Foreman & Katello, but I'd prefer to deal with Candlepin & Pulpcore in the same commit.

theforeman.org/pipelines/vars/candlepin/4.3.groovy
```diff
@@ -3,7 +3,7 @@ def candlepin_distros = ['el8', 'el9']
 def packaging_branch = 'rpm/4.3'
 def pipelines = [
     'candlepin': [
-        'centos8-stream',
-        'centos9-stream'
+        'centos9-stream',
+        'almalinux8'
```
ekohl, line 7: It wasn't done before, but I prefer trailing commas so you can easily add other things later.

ekohl (changes requested):
Please add `almalinux9` where you added `almalinux8` as well. That way we're consistently testing it in the same way everywhere.

ekohl (approved):
Thanks!

PR #479, opened 2024-06-05
use correct vars for pulpcore repoclosure

evgeni:
Fixes: cdfad4759b0f0e17cc8509b275fdcb34f8ff9bb1


ekohl (approved):
I noticed the error yesterday, but didn't come up with a patch yet. Thanks!

PR #482, opened 2024-07-12
use main instead of master for hammer-cli-katello

evgeni:


evgeni:
cc @ianballou 

should we also rename the jobs to say `main`? (will require some changes in packaging too then)

evgeni:
:woman_shrugging: 

I only update to match https://github.com/Katello/hammer-cli-katello/branches and thus make things work again.

ehelms (approved):
@ekohl I think `main` was chosen because it is what Github changed its default to. Your comment around `develop` is fair as an ecosystem decision but it's something we've never fully articulated as a thing we should standardize around. I think this could be worth a discourse post and update to developer docs.

evgeni:
> @ekohl I think `main` was chosen because it is what Github changed its default to. Your comment around `develop` is fair as an ecosystem decision but it's something we've never fully articulated as a thing we should standardize around. I think this could be worth a discourse post and update to developer docs.

Ian did ask on Matrix some time ago, and we recommended `develop` for consistency ;)

PR #489, opened 2024-08-21
Add 3.12 stable

pcreech:


ehelms (approved):

ekohl (approved):
I noticed there's also no 3.12 view. Would you mind looking at that too?

PR #492, opened 2024-09-11
explicitly migrate the test DB

evgeni:


theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -20,7 +20,7 @@ def configureDatabase(ruby) {
     bundleExec(ruby, 'rake db:drop >/dev/null 2>/dev/null || true')
     bundleExec(ruby, 'rake db:create --trace')
     bundleExec(ruby, 'rake db:create --trace RAILS_ENV=production')
-    bundleExec(ruby, 'rake db:migrate --trace')
```
ekohl, line 23: I'd like consistency with the `db:create` statements.

evgeni: and by consistency you mean "delete the creates, they are implicit by migrate"? :)

evgeni: mhh, we still need `create` for prod (which is not migrated, but needs to exist for asset compilation, and we probably do not properly set `DATABASE_URL=nulldb://nohost` everywhere)

evgeni: so I updated the create to have `RAILS_ENV=test`, that should be safe, we can get things rolling (again) and do the cleanup later.

ekohl (approved):

PR #494, opened 2024-10-07
Drop EL8 from Pulpcore RPM Packaging

Odilhao:


ekohl (approved):
Looks good once we've completed the EL8 removal with Foreman (technically Katello, but we'll combine those).

Odilhao:
We can merge this one now.

PR #495, opened 2024-10-09
Drop EL8 from Foreman, Katello and Candlepin

Odilhao:


theforeman.org/pipelines/vars/foreman/nightly.groovy
```diff
@@ -7,8 +7,7 @@ def foreman_client_distros = [
     'el7'
 ]
 def foreman_el_releases = [
-    'el9',
-    'el8'
```
evgeni, line 11: The pipelines, they are still running

evgeni (changes requested):

theforeman.org/pipelines/vars/foreman/nightly.groovy
evgeni: Sure, but this PR doesn't disable them ;)

ekohl: Ah, you mean that they should also be dropped from the `centos.org` directory?

evgeni: At least from `def pipelines_el ` in this file :)

Odilhao:
Done removed the el8 from both install and upgrade pipelines.

evgeni (approved):

PR #497, opened 2024-10-23
enable upgrade tests for Debian 12

evgeni:
we actually have packages for 3.11 and 3.12, so we can test nightly
upgrades too

### DIFF
--- a/theforeman.org/pipelines/vars/foreman/nightly.groovy
+++ b/theforeman.org/pipelines/vars/foreman/nightly.groovy
@@ -20,6 +20,7 @@ def pipelines_deb = [
     ],
     'upgrade': [
         'debian11',
+        'debian12',
         'ubuntu2204'
     ]
 ]


ekohl (approved):
Do we also want to enable Debian 12 in the 3.11 & 3.12 pipelines?

evgeni:
> Do we also want to enable Debian 12 in the 3.11 & 3.12 pipelines?

That's something someone with a release hat needs to decide. ;)
Technically 3.11 should install fine, and starting with 3.12 we also have a possible upgrade path.

(I have tested nightly install and upgrade pipes, so quite confident "it would just work")

ekohl:
With a release hat on, I'd like to add them. We're doing a 3.11.4 release as we speak and if we add it to the pipelines then I'd be confident in including a release note that we now support it with Debian 12, encouraging users to plan their upgrade.

evgeni:
https://github.com/theforeman/jenkins-jobs/pull/498

PR #500, opened 2024-11-05
Foreman 3.13 branching

pcreech:


theforeman.org/yaml/jobs/test_3_13_stable.yaml
```diff
@@ -0,0 +1,31 @@
+- job:
+    name: test_3_12_stable
```
ekohl, line 2:
```suggestion
    name: test_3_13_stable
```

ekohl (changes requested):

theforeman.org/yaml/jobs/test_3_13_stable.yaml
pcreech: should be good now

ekohl (approved):

PR #510, opened 2024-11-28
deploy website to website01

evgeni:


ekohl (commented):
I've been wondering about using the GitHub deployments API as well. Mostly because there's a more direct view of what was deployed in our PRs. Have you used https://docs.github.com/en/actions/use-cases-and-examples/deploying/deploying-with-github-actions in the past?

Also, it made me look at this file and noticed we made some preparations for the CDN but didn't activate it when we turned on the CDN:
https://github.com/theforeman/jenkins-jobs/blob/4a0b9c2533d48a8a3085908c808975354fde12c6/theforeman.org/pipelines/deploy/website.groovy#L35-L38
Should we do that now?

evgeni:
> I've been wondering about using the GitHub deployments API as well. Mostly because there's a more direct view of what was deployed in our PRs. Have you used https://docs.github.com/en/actions/use-cases-and-examples/deploying/deploying-with-github-actions in the past?

I did not (and didn't plan to do that overhaul right now)

> Also, it made me look at this file and noticed we made some preparations for the CDN but didn't activate it when we turned on the CDN:
> 
> https://github.com/theforeman/jenkins-jobs/blob/4a0b9c2533d48a8a3085908c808975354fde12c6/theforeman.org/pipelines/deploy/website.groovy#L35-L38
> 
> Should we do that now?

That's unrelated to which backend the CDN uses, but yes.

evgeni:
https://github.com/theforeman/jenkins-jobs/pull/512

ekohl:
> I did not (and didn't plan to do that overhaul right now)

I took a stab at it in https://github.com/theforeman/theforeman.org/pull/2206. What we could even do is to have both in place for a time: GHA for the new host and Jenkins for the old host.

ekohl (approved):

PR #511, opened 2024-11-28
deploy downloads to website01

evgeni:


ekohl (commented):
Why is the hostname website01? I'd have thought web02 would be more logical. My mind thought it was short for webserver

evgeni:
> Why is the hostname website01? I'd have thought web02 would be more logical. My mind thought it was short for webserver

Idk what I thought when I created it, but it's there now.

evgeni:
Oooh, probably because I didn't want it to match `web\d+` that's used to apply the *big* web class/profile

ekohl:
> Oooh, probably because I didn't want it to match `web\d+` that's used to apply the _big_ web class/profile

That's fair. It's indeed not an exact drop in.

ekohl (approved):

PR #514, opened 2024-11-28
Use rpm context for packaging PRs

ekohl:
This uses the context rpm again instead of rpm-copr. Now that koji is gone, there's no more need for this distinction.


ehelms (approved):

ekohl:
@Odilhao @pcreech @zjhuntin for visibility: you need to use `[test rpm]` from now on, instead of `[test rpm-copr]`.

evgeni:
you should have updated the branch rules too (done) :)

PR #523, opened 2025-01-06
Rewrite Foreman's testing to a matrix

ekohl:
This makes it easier to change the Ruby versions and also test on multiple at the same time.

### DIFF
--- a/theforeman.org/pipelines/release/source/foreman.groovy
+++ b/theforeman.org/pipelines/release/source/foreman.groovy
@@ -1,5 +1,5 @@
 pipeline {
-    agent { label 'fast' }
+    agent none
 
     options {
         timestamps()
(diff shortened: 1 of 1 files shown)


ekohl:
@ehelms @evgeni my goal is to also replace the old `test_X_Y_stable` jobs with this pipeline so I made it a bit more generic. On the other hand, I still struggle with the testing matrix that should be unique for every version.

When I look at `theforeman.org/pipelines/release/source/katello.groovy` I see we set a single variable and pick a single version. Should we do that too for Foreman to keep it simple?

theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,3 +1,22 @@
+def railsEnvForTask(task, env) {
```
evgeni, line 1:
```suggestion
def databaseUrlForTask(task, env) {
```

theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,3 +1,22 @@
...
+        task_id = task.split(':')[1]
+        database = (['test', env.JOB_NAME, env.BUILD_ID] + ruby + [task_id]).join('-')
+        if (database.size() > 63) {
+            error "${id} cannot be used to generate a PostgreSQL DB name, the resulting name would be longer than 63 chars."
+        }
+        return "postgresql://foreman:foreman@localhost/${database}"
+    }
+}
+
+def databaseUrlForTask(task) {
```
evgeni, line 16:
```suggestion
def railsEnvForTask(task) {
```

theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,3 +1,22 @@
...
+    return task == 'assets:precompile' ? 'production' : 'test'
+}
+
 def databaseFile(id) {
```
evgeni, line 20: is this still used somewhere then?

ekohl: https://github.com/theforeman/jenkins-jobs/blob/e4c55790f55bfc196a5b7eda5460b7e20ce37818/theforeman.org/pipelines/release/source/katello.groovy#L30
Do you want me to also replace that?

evgeni: `theforeman.org/pipelines/release/source/katello.groovy` does use this, but I guess it'd get the same treatment once this PR is done and we see it working fine

ekohl: Unrelated, but I had a closer look at the other functions and noticed `addSettings` was also unused: https://github.com/theforeman/jenkins-jobs/pull/524

theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -8,98 +8,61 @@ pipeline {
...
-                stage('ruby-2.7-postgres') {
-                    agent { label 'fast' }
-                    environment {
-                        RUBY_VER = '2.7.6'
+            matrix {
+                agent { label 'fast' }
+                axes {
+                    axis {
+                        name 'RUBY_VER'
+                        values '2.7.6'
```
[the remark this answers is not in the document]
ekohl: That's indeed my struggle. AFAIK you can't use a variable here and need to hardcode it. Then you could use filtering to bring it down, but that feels so complicated.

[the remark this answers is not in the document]
evgeni: We have a couple of `parallel <weird_Map_that_defines_the_branches>` in our repo, and I hate it… :+1: for not doing that here, TBH

theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -8,98 +8,61 @@ pipeline {
         buildDiscarder(logRotator(daysToKeepStr: '7'))
     }
 
+    environment {
+        BUNDLE_WITHOUT = 'development'
+        RAILS_ENV = railsEnvForTask(TASK)
+        DATABASE_URL = databaseUrlForTask(TASK, env)
```
ekohl, line 14: I suspect you're not allowed to pass in `env` here:
```
java.lang.IllegalArgumentException: One or more variables have some issues with their values: DATABASE_URL
```

Thoughts on just using some UUID for the DB name?

evgeni (approved):

evgeni:
Mhh, https://ci.theforeman.org/job/foreman-develop-source-release/ is failing, and the number of tests is different for the passed jobs? :thinking:

theforeman.org/pipelines/lib/foreman.groovy
```diff
@@ -1,3 +1,16 @@
+def databaseUrlForTask(task) {
+    if (task == 'assets:precompile') {
+        return 'nulldb://nohost'
+    } else {
+        database = UUID.randomUUID().toString()
+        return "postgresql://foreman:foreman@localhost/foreman-${database}"
```
evgeni, line 6: somehow all tests end up using the same DB. and that breaks things

evgeni (changes requested):

theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -8,98 +8,64 @@ pipeline {
         buildDiscarder(logRotator(daysToKeepStr: '7'))
     }
 
+    environment {
+        RUBY_VERSION = '2.7.6'
+        BUNDLE_WITHOUT = 'development'
+        RAILS_ENV = railsEnvForTask(RAKE_TASK)
+        DATABASE_URL = databaseUrlForTask(RAKE_TASK)
```
evgeni, line 15: I think this gets evaluated too early?

theforeman.org/pipelines/lib/foreman.groovy
ehelms: Is this still true with the changes that moved the function call into the matrix?

ekohl: Tested in https://ci.theforeman.org/blue/organizations/jenkins/ewoud-test/detail/ewoud-test/2/pipeline/77 (likely to be deleted) but there I see:

> [2025-03-06T13:45:30.061Z] Created database 'foreman-c85b58bc-19d6-453b-8e06-baa2f933dae5'
> [2025-03-06T13:44:23.509Z] Created database 'foreman-7d94bab3-8303-425f-96a3-d7d8db402c5b'

So moving it into the matrix helps.

ekohl:
Updated now that https://github.com/theforeman/jenkins-jobs/pull/525 has been merged. I've tried to keep the diff minimal.

One thing to note is that we should keep Ruby on the oldest version we support, not the newest. Otherwise our packaging job may pull in too new gems that no longer support the old version.

ehelms (approved):

theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -1,5 +1,5 @@
 pipeline {
-    agent { label 'fast' }
+    agent any
```
evgeni, line 2: should this be `none`, like `theforeman.org/pipelines/test/foreman.groovy`, or is `any` required for `post` to work properly?

(Reading it again, "Build and Archive Source" also has no agent defined)

theforeman.org/pipelines/release/source/foreman.groovy
```diff
@@ -110,7 +84,7 @@ pipeline {
                     git url: git_url, branch: git_ref
```
evgeni, line 84: one thing I realized (but it's not a regression): `git_ref` is `develop` here, right?
So the tests can run with `develop (abcdef12)`, but the source generation could run with `develop (dedede12)` if someone pushed to `develop` in-between.

evgeni (approved):

theforeman.org/pipelines/release/source/foreman.groovy
ekohl: It was my guess that it was needed for the source build

theforeman.org/pipelines/release/source/foreman.groovy
ekohl: I think I've seen some weirdness in that area before. Probably something to look into with a follow up

theforeman.org/pipelines/release/source/foreman.groovy
evgeni: I think I'd prefer a `none` here and an explicit `any` in the source build, but not blocking :)

ekohl: I looked at it but wasn't sure you could set it for an individual stage after reading the docs. I can take another look

evgeni: https://github.com/theforeman/jenkins-jobs/blob/2b2129bada42ef4ce1d68881d2f007ccbf7cb737/theforeman.org/pipelines/release/source/foreman.groovy#L77-L78

we do that today? :)

ekohl: Done. I also moved the `cleanup` to the specific stage because I think that will work better. I don't think the general notification steps in `post` need an agent.

ekohl:
Shall we merge this?

evgeni:
Yeah, I think it's a good time for a bit of CI breakage :)

ehelms:
All ahead full for merge.

PR #525, opened 2025-01-07
Rewrite Foreman's stable job testing to a pipeline

ekohl:
This takes inspiration from https://github.com/theforeman/jenkins-jobs/pull/523 and makes life better by being consistent. That consistency also removes a step in the branching procedure (and clean up later on when a version is dropped).


theforeman.org/pipelines/test/foreman.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+    environment {
+        RUBY_VERSION = ruby_version
+        BUNDLE_WITHOUT = 'development'
+        RAILS_ENV = railsEnvForTask(RAKE_TASK)
+        DATABASE_URL = databaseUrlForTask(RAKE_TASK)
+        TESTOPTS = '-v'
+    }
+
+    matrix {
```
ehelms, line 16: I think this has to be a child of at least a stages-stage declaration:

 ```    
  stages {
      stage('') {
        matrix { 
```

File: theforeman.org/pipelines/test/foreman.groovy [hunk 2]
```diff
@@ -0,0 +1,64 @@
+pipeline {
```
ehelms, line 1: I think tihs top level pipeline needs an agent declaration:

```
pipeline {
  agent none
```

ekohl:
It now runs, but I see this weird failure:
```
[2025-03-06T13:54:00.322Z] Index name 'index_messages_on_value' on table 'messages' does not exist
[2025-03-06T13:54:00.322Z] /home/jenkins/.rubygems/ruby/2.7.0/gems/activerecord-6.1.7.10/lib/active_record/migration/compatibility.rb:290:in `index_name_for_remove'
[2025-03-06T13:54:00.322Z] /home/jenkins/.rubygems/ruby/2.7.0/gems/activerecord-6.1.7.10/lib/active_record/migration/compatibility.rb:265:in `remove_index'
[2025-03-06T13:54:00.322Z] /home/jenkins/workspace/ewoud-test@2/db/migrate/20130121130826_add_digest_to_messages.rb:3:in `up'
```

This is within the nulldb test, but I wonder why this only shows up now.

ehelms (approved):
Looks good from a code perspective, and having addressed the uniqueness problem from what I can see. I did not run it.

ekohl:
https://github.com/theforeman/theforeman-rel-eng/pull/474 will update the rel eng scripts.

@pcreech @Odilhao is it OK to break the stable pipelines now? ;)

pcreech:
We still have Foreman 3.14 and Katello 4.16 GA to push.

If you can fix any breakages before then, lets GO.  

If not, might be better to wait.

(I'm a chaos gremlin, so, I'd yolo merge personally, but y'all might be a little more cautious than I am)

Odilhao (approved):

evgeni:
given 3.14/4.16 GAs were pushed out yesterday, I think we're good to go here?

ekohl:
Technically it's also just a sanity validation and shouldn't block a 3.16.1 release (if we need it) so I'd be happy to merge this. Next week I won't be around to fix things that I'll likely have broken. If @pcreech and @Odilhao are happy to step in then let's go for it.

evgeni:
:woman_shrugging:

PR #526, opened 2025-01-08
drop smart-proxy-pr-test and all non-packaging PR testing remains

evgeni:


evgeni:
We have GHA now: https://github.com/theforeman/smart-proxy/pull/909

theforeman.org/pipelines/test/smart-proxy.groovy
```diff
@@ -28,9 +20,7 @@ pipeline {
             stages {
                 stage('Setup Git Repos') {
                     steps {
-                        script {
-                            git_checkout()
-                        }
+                        git branch: "${params.BRANCH}", url: 'https://github.com/theforeman/smart-proxy'
```
ekohl, line 23: I think we can reduce it even further since the params are just a hack to easily get the variable in scope. Perhaps it was never the best solution, but it's what's here now:
https://github.com/theforeman/jenkins-jobs/blob/140034708fc72e5c36466373f62c60335768d014/theforeman.org/pipelines/lib/stable/smart-proxy.groovy#L4

So it can be:
```suggestion
                        git branch: git_branch, url: 'https://github.com/theforeman/smart-proxy'
```

And then you can drop `lib/stable/smart-proxy.groovy` altogether

ekohl (approved):

PR #529, opened 2025-01-08
Align the develop and stable testing jobs

ekohl:
This is a bit of an experiment to see if we can use the same jobs for testing develop and stable branches. It's inspired by https://github.com/theforeman/jenkins-jobs/pull/525.

Some differences we should discuss:
* It still runs RuboCop on develop, but should we? it's not something we care about
* It only builds the package for develop, but perhaps we can just build it for all
* Build rotation: can we align on a week of builds and X builds? Not sure
* Shall we drop the matrix building and align on a single Ruby version per release? Do we pick the oldest version?

### DIFF
--- a/theforeman.org/pipelines/release/source/smart-proxy.groovy
+++ b/theforeman.org/pipelines/release/source/smart-proxy.groovy
@@ -1,9 +1,9 @@
 pipeline {
-    agent any
+    agent none
 
     options {
         timestamps()
-        timeout(time: 2, unit: 'HOURS')
+        timeout(time: 1, unit: 'HOURS')
         ansiColor('xterm')
         buildDiscarder(logRotator(daysToKeepStr: '7'))
     }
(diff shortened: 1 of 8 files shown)


ekohl:
@ehelms @evgeni this is IMHO the next step to simplify things in our CI. I'd appreciate your thoughts.

theforeman.org/pipelines/release/source/smart-proxy.groovy
```diff
@@ -58,37 +74,17 @@ pipeline {
...
-            deleteDir()
-        }
-    }
-}
-
-def run_test(args) {
-    def ruby = args.ruby
-
-    try {
-        sh "cp config/settings.yml.example config/settings.yml"
```
evgeni, line 87: isn't this step now missing?

ekohl: https://github.com/theforeman/smart-proxy/blob/develop/config/settings.yml.example only contains commented lines so it shouldn't make a difference. `theforeman.org/pipelines/test/smart-proxy.groovy` also doesn't do it so I'm assuming the unit test suite doesn't need it.

Perhaps we should enhance the smart proxy to deal with running without a config where currently it can't?

evgeni: Wfm

evgeni (approved):

PR #533, opened 2025-02-13
Foreman 3.14 branching

pcreech:


Odilhao (approved):

ekohl (changes requested):
You missed https://github.com/theforeman/jenkins-jobs/blob/master/theforeman.org/yaml/views/release.yml so we wouldn't have a corresponding view.

ekohl (approved):

PR #536, opened 2025-03-07
Use Python 3.12 to install Ansible dependencies on Duffy

evgeni:


evgeni:
merging myself to unblock things

PR #549, opened 2025-08-08
branching/deprecate old/3.16

ogajduse:
This PR depends on #548 being merged first.


ogajduse:
Switching to ready state as 3.16 is GA.

zjhuntin (approved):

PR #567, opened 2025-12-16
Run foreman nightly rpm pipeline earlier in the day.

zjhuntin:


ekohl:
What's the motivation for this change?

ogajduse:
We are shifting the nightly pipeline schedule to better align with our primary time zones (ET and CET). By moving the run to a window after the ET workday ends but before the CET workday begins, we ensure that all daily contributions are captured.

The current 21:00 CET (15:00 ET) slot often misses late-afternoon changes from the ET team. This update allows ET developers to push their final updates and releases with the confidence that they will be tested overnight and ready for review the following morning.

ekohl (approved):
Thanks! That explanation helps

ogajduse (approved):

PR #578, opened 2026-06-04
Isolate katello rubocop bundle to fix Rack version conflicts

zjhuntin:
## Summary
- The rubocop stage ran a separate `bundleInstall` from the katello root, which resolved Rails 7.1 / Rack 3 (katello.gemspec has no Rails version constraint). This ran in parallel with the test stage using Foreman's bundle (Rails 7.0 / Rack 2), both sharing `~/.rubygems`, causing VCR body_json matcher failures in the Pulp3 MultiCopyAllUnit tests.
- Isolate katello's rubocop bundle via `BUNDLE_PATH` so it no longer conflicts with Foreman's parallel bundle at `~/.rubygems`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/theforeman.org/pipelines/release/source/katello.groovy
+++ b/theforeman.org/pipelines/release/source/katello.groovy
@@ -64,9 +64,8 @@ pipeline {
                         BUNDLE_PATH = "${env.WORKSPACE}/.rubocop_bundle"
                     }
                     steps {
-                        dir('foreman') {
-                            bundleExec(ruby, "rubocop --parallel ../${project_name}")
-                        }
+                        bundleInstall(ruby)
+                        bundleExec(ruby, 'rubocop --parallel')
                     }
                 }
                 stage('react-ui') {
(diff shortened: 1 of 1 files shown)


ekohl (changes requested):
This is not how CI runs it:
https://github.com/Katello/katello/blob/b1c394a56d4d494dcc81132606aa87779ec9dcb4/.github/workflows/ruby.yml#L16-L21
This is does a checkout of `katello/katello` and installs it from there (https://github.com/Katello/katello/blob/b1c394a56d4d494dcc81132606aa87779ec9dcb4/Gemfile#L9).

ekohl (commented):
It surprises me this is needed, but if concurrent access to the shared Rubygems directory is a problem then this can help.

Should it also revert https://github.com/theforeman/jenkins-jobs/pull/577 and align both approaches?

ekohl (approved):
Code wise 👍, but untested. Could you also update the PR title to match the content?

PR #579, opened 2026-06-10
Add Debian 15 and Ubuntu 26.04 codenames to packaging config

evgeni:


theforeman.org/pipelines/lib/packaging.groovy
```diff
@@ -429,9 +429,11 @@ def inject_debian_release_version(os) {
         'bookworm': 'debian12',
         'trixie': 'debian13',
         'forky': 'debian14',
+        'duke': 'debian15',
```
devendra104, line 432: just curious to know, since duke is not yet out, what is the need for this change?

evgeni: It's not needed (yet), but as I was updating the list for 26.04 anyway (which is needed), and knew the name for Debian 15, I added it :)

Source: https://www.debian.org/releases/

devendra104 (approved):

PR #584, opened 2026-07-06
Retrigger Konflux nightly component rebuilds after katello nightly pipeline

Odilhao:
## Summary
- Adds a `trigger-konflux-rebuild` stage to `katello-nightly-rpm-pipeline` (`pipelines/release/pipelines/katello.groovy`), right after `staging-push-rpms`, that retriggers Konflux rebuilds for the four nightly (develop) components (`candlepin-develop`, `foreman-develop`, `foreman-proxy-develop`, `pulp-develop`) by annotating their Component CRs (`build.appstudio.openshift.io/request: trigger-rebuild`), per the [Konflux API retrigger docs](https://konflux-ci.dev/docs/building/running/#retriggering-a-post-merge-build-from-your-main-branch-from-the-api).
- New `pipelines/lib/konflux.groovy` downloads the `oc` client at runtime (stable channel, removed after use), logs in with a masked `string` credential binding (`konflux-jenkins-trigger-token`, wrapping the login line in `set +x`/`set -x` so the token never hits the console log), then runs `oc annotate` for each component.
- Scoped to katello nightly only for now. The 3.19 variants and `foreman-mcp-server-develop` will be added separately once this is proven.

## Where it's wired in and why
- It's a regular pipeline `stage`, not a `post { success }` block. `post` blocks are unreliable with Kubernetes/OpenShift pod agents (the pod backing the agent can already be torn down by the time `post` runs), so the retrigger runs as the last stage instead - it's naturally skipped if an earlier stage fails, same net effect as `post { success }` without that risk.
- It runs after the full `staging-build-repository` → `staging-copy-repository` → `staging-repoclosure` → `staging-test` → `staging-push-rpms` chain, i.e. only once the nightly RPMs have actually been staged, tested via `runDuffyPipeline`, and pushed - not right after the raw Copr/DEB build in `katello-master-package-release`. That job only builds packages, so triggering off it would rebuild Konflux images from packages that might still fail staging tests.
- `konflux_components` is defined in `pipelines/vars/katello/nightly.groovy`, which `katello-nightly-rpm-pipeline` already includes. `pipelines/release/pipelines/katello.groovy` is shared with the stable `katello-{version}-rpm-pipeline` jobs (4.20, 4.21), which use per-version vars files that don't define `konflux_components` - the stage's `when` guards on `binding.hasVariable('konflux_components')` so those jobs are unaffected.

## Discussion point for reviewers
Putting `konflux_components` in `pipelines/vars/katello/nightly.groovy` ties image rebuild triggering directly to the katello nightly pipeline's vars. That works for now, but two things make me want a second opinion on placement before we lean on it further:
- We're about to branch this (3.19 variants, `foreman-mcp-server-develop`), and those may not all map 1:1 to a single project's nightly vars file the way `candlepin-develop`/`pulp-develop` currently piggyback on katello's.
- Conceptually, Konflux image rebuilds aren't really "a katello pipeline concern" - e.g. an image containing only foreman (no katello) would have no natural home in `katello/nightly.groovy`.

Open to alternatives, e.g. a dedicated `pipelines/vars/konflux/nightly-components.groovy` (or similar) that's included explicitly by whichever pipeline(s) need to fire triggers, decoupling "which components rebuild" from "which project's nightly vars happen to run last."

## Test plan
- [x] `./test theforeman.org` passes locally (all jobs OK, including `katello-nightly-rpm-pipeline`, `katello-4.20-rpm-pipeline`, `katello-4.21-rpm-pipeline`, and `katello-master-package-release`, via the real declarative-pipeline validator against `ci.theforeman.org`)
- [ ] Confirm on a real run that the four `-develop` Konflux components get annotated and a rebuild is queued after a successful nightly pipeline

### DIFF
--- a/theforeman.org/pipelines/lib/konflux.groovy
+++ b/theforeman.org/pipelines/lib/konflux.groovy
@@ -1,0 +1,43 @@
+def konflux_api_server = 'https://api.kflux-fedora-01.84db.p1.openshiftapps.com:6443'
+def konflux_namespace = 'theforeman-org-tenant'
+def konflux_oc_channel = 'stable'
+
+def retrigger_konflux_components(components) {
+    if (!components) {
+        return
+    }
+
+    def oc_dir = "${env.WORKSPACE}/bin"
+    def oc_bin = "${oc_dir}/oc"
+
+    sh(
+        label: 'download oc client',
+        script: """
+            mkdir -p ${oc_dir}
+            curl -fsSL --retry 3 https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${konflux_oc_channel}/openshift-client-linux.tar.gz | tar -xz -C ${oc_dir} oc
+            chmod +x ${oc_bin}
+        """
+    )
+
+    try {
+        withCredentials([string(credentialsId: 'konflux-jenkins-trigger-token', variable: 'KONFLUX_TOKEN')]) {
+            sh(
+                label: 'oc login to konflux (token hidden)',
+                script: """
+                    set +x
+                    ${oc_bin} login --token=\$KONFLUX_TOKEN --server=${konflux_api_server}
+                    set -x
+                """
+            )
+        }
+
+        components.each { component ->
+            sh(
+                label: "trigger konflux rebuild: ${component}",
+                script: "${oc_bin} annotate component ${component} --overwrite --namespace ${konflux_namespace} build.appstudio.openshift.io/request=trigger-rebuild"
+            )
+        }
+    } finally {
+        sh(label: 'remove oc client', script: "rm -rf ${oc_dir}")
+    }
+}
--- a/theforeman.org/pipelines/lib/konflux.groovy
+++ b/theforeman.org/pipelines/lib/konflux.groovy
@@ -9,6 +9,7 @@ def retrigger_konflux_components(components) {
 
     def oc_dir = "${env.WORKSPACE}/bin"
     def oc_bin = "${oc_dir}/oc"
+    def kubeconfig = "${oc_dir}/kubeconfig"
 
     sh(
         label: 'download oc client',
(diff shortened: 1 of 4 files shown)


theforeman.org/pipelines/lib/konflux.groovy
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        components.each { component ->
+            sh(
+                label: "trigger konflux rebuild: ${component}",
+                script: "${oc_bin} annotate component ${component} --overwrite --namespace ${konflux_namespace} build.appstudio.openshift.io/request=trigger-rebuild"
+            )
+        }
+    } finally {
+        sh(label: 'remove oc client', script: "rm -rf ${oc_dir}")
```
ogajduse: Woukd it make sense to `oc logout` here too?

Odilhao: if the runs are ephemeral I don't see why, unless the workspace is not cleaned after the execution.

zjhuntin (commented):
## Review: Retrigger Konflux nightly component rebuilds after katello nightly pipeline

**Scope**: 4 files changed, +56 lines | Adds a new pipeline stage and shared library to retrigger Konflux component rebuilds after successful katello nightly RPM staging

---

### Summary

Adds a `trigger-konflux-rebuild` stage at the end of the katello nightly RPM pipeline that downloads the `oc` CLI, logs into the Konflux cluster, and annotates four nightly Component CRs to trigger image rebuilds. The new logic is cleanly isolated in `pipelines/lib/konflux.groovy` and guarded with `binding.hasVariable('konflux_components')` so stable pipelines (4.20, 4.21) are unaffected. Well-structured PR with excellent description explaining placement rationale.

### Verdict: APPROVE (with minor suggestions)

---

### Important Issues (should fix)

- **`konflux.groovy:28`**: The `oc login` runs inside `withCredentials` but the `oc annotate` calls at line 34-38 run *outside* that block. `oc login` writes a kubeconfig/token file to `~/.kube/config` (or `$KUBECONFIG`), so the token persists on the agent's filesystem after `withCredentials` closes. This works, but the token is sitting in the kubeconfig on the agent until `rm -rf ${oc_dir}` in the `finally` — and that only deletes the `oc` binary, not the kubeconfig. Consider either:
  - Moving the `components.each` loop inside the `withCredentials` block (simplest), or
  - Adding `oc logout` or explicitly cleaning up `~/.kube/config` in the `finally` block (this is what @ogajduse raised in the inline comment too)
  - **Why it matters**: If the pipeline fails between login and the `finally` block cleanup, the token remains on the agent. The `el`-labeled agents may be shared.

### Minor Suggestions (nice to have)

- **`konflux.groovy:17`**: The `curl | tar` downloads the full stable OCP client on every nightly run. This is fine for now, but if the mirror is ever slow or flaky it'll block the pipeline. A pinned version (e.g. `stable-4.17` instead of `stable`) would make builds more reproducible and avoid surprise breakage from new `oc` releases.

- **`konflux.groovy:1-3`**: The `konflux_api_server`, `konflux_namespace`, and `konflux_oc_channel` are hardcoded as script-level variables. This works since the file is a shared library, but if a second pipeline ever needs different values, these would need to become parameters. Fine for now — just noting for the "where does config live" discussion the author raised.

### Positive Observations

- Excellent PR description — the reasoning for placing this as a `stage` rather than `post { success }` (pod agent teardown risk) is exactly right for Jenkins declarative pipelines on Kubernetes agents.
- Clean separation: the shared library in `pipelines/lib/konflux.groovy` is reusable and the vars-based component list makes it easy to extend.
- The `try/finally` cleanup of the `oc` binary is good practice.
- The `set +x` / `set -x` around the token-bearing login command is the correct Jenkins pattern for masking secrets in shell steps.
- The YAML change to `katello-nightly-rpm-pipeline.yaml` correctly adds the new lib include — stable pipelines don't include it and don't need it.

---

### Questions for the Author

- On the discussion point about placement of `konflux_components`: the author's suggestion of a dedicated `pipelines/vars/konflux/nightly-components.groovy` seems like the right direction once this expands beyond katello nightly. Is there a timeline for the 3.19 variants / `foreman-mcp-server-develop` additions that would force that refactor?
- The credential ID `konflux-jenkins-trigger-token` — is this already provisioned on `ci.theforeman.org`, or does that need to happen before this can merge and run?

theforeman.org/pipelines/lib/konflux.groovy
Odilhao: You're right, and I need to correct my earlier reply — this pipeline runs on `agent { label 'el' }`, a fixed pool of persistent machines, not the ephemeral Cloud/Kubernetes pod agents we use elsewhere. The workspace and a default `~/.kube/config` can persist across builds on the same runner, so the token would sit there.

Fixed by scoping `KUBECONFIG` to a file under the stage's own workspace dir (`${oc_dir}/kubeconfig`) instead of the default `~/.kube/config`, and adding an explicit `oc logout` before the existing `rm -rf ${oc_dir}` cleanup in `finally`. That way the token never touches shared state on the runner and is wiped out on both the happy and failure paths.

zjhuntin (approved):

ogajduse (approved):

PR #571, opened 2026-05-07
Remove centos.org and integrate Duffy testing into theforeman.org

zjhuntin:
## Summary

Incremental approach to eliminate the two-Jenkins model. Starting with a single consolidated job rather than refactoring everything at once (following Ewoud's suggestion in review).

### Phase 1: Consolidate foreman-discovery-image (IN PROGRESS)
- ✅ Create `foreman-discovery-image-build` job on theforeman.org with direct Duffy provisioning
- ✅ Update `foreman-discovery-image-release` to call the new build job
- ⏳ Test in Jenkins and iterate

### Phase 2: Additional jobs (planned)
Once the foreman-discovery-image job is validated, we can consolidate other jobs and eventually:
- Move ansible playbooks from `centos.org/` to `theforeman.org/ansible/`
- Add `duffy_test.groovy` library with `runDuffyTest()` for direct Duffy test execution
- Rewrite `foreman_infra.groovy` to call Duffy directly
- Add single parameterized `duffy-integration-test` job replacing ~83 generated centos.org jobs
- Update `duffy.groovy` paths, branch scripts, test script, and README
- Remove `ci-centos-integration-test` job
- Delete entire `centos.org/` directory

Fixes https://github.com/theforeman/foreman-infra/issues/1828

## How to review/test

- Run `./test theforeman.org` to confirm JJB generates valid XML
- Check the new `foreman-discovery-image-build` job logic (moved from centos.org)
- Once merged, watch the first FDI nightly build — verify Duffy provisioning and publishing work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #588, opened 2026-07-17
Add konflux-gate-pipeline: stage-test and promote Konflux images

Odilhao:
## Summary

Konflux nightly builds for `candlepin-develop`, `foreman-develop`,
`foreman-proxy-develop` and `pulp-develop` currently auto-release straight to
production quay with no functional testing. This adds `konflux-gate-nightly-pipeline`,
chained after `katello-nightly-rpm-pipeline`, which:

1. Waits for each Konflux Application's Snapshot to update (per-component timeout,
   falls back to the last successfully-released Snapshot on timeout so one broken
   component doesn't block the others).
2. Runs a real `foremanctl`/`forge` install test against the staged images
   (`quay.io/foreman/stage/*`), provisioned directly via Duffy on `ci.theforeman.org`.
3. On pass, creates a Konflux `Release` CR per Application to promote stage → production
   quay, polls it to completion, and records the actual published digest in the build
   description.

## Design decisions

- **Duffy directly on ci.theforeman.org, not delegated through ci.centos.org.** The
  existing `runCicoJob`/`runIndividualCicoJob` path HTTP-triggers a job on the
  centos.org Jenkins for every Duffy-backed test today — an extra hop/failure domain
  not needed here. Draft PR #571 is already moving Duffy test-driving into
  theforeman.org; this follows that direction (`provisionDuffy()`/`duffy_ssh()` drive a
  single `root@duffy_box` SSH session, modeled on
  `centos.org/pipelines/foreman-discovery-image-build.groovy`, not forklift's
  multi-user abstraction). No new centos.org delegation was added.
- **foremanctl/forge, not forklift.** `forge` is what `foreman-oci-images`' own CI
  already uses for this install test. We reuse its VM lifecycle and deploy invocation
  as-is but skip the local `podman build`/`save`/`load` step — Konflux already built and
  pushed the images we're gating, so we only ever pull the real digest-pinned
  `quay.io/foreman/stage/*` refs (see reproduction below). `forge` itself transitively
  pulls the `theforeman.forklift` Ansible collection to manage the VM — that's `forge`'s
  own dependency, not one this PR adds.
- **No fixed-delay/tkn polling for the rebuild trigger.** `katello.groovy` captures each
  Application's "previous" Snapshot *before* triggering the Konflux rebuild
  (`capture-konflux-snapshots`) and passes it to the gate job as a build parameter
  (`PREVIOUS_SNAPSHOTS`) rather than letting the gate job recompute it after starting —
  recomputing there would race with Konflux's own build latency (a fast build could
  already have landed by the time the gate job's first stage runs). Comparing against
  the snapshot name itself already blocks for exactly as long as the real build takes.
- **Release CRs are polled to completion, not fire-and-forget.** Konflux's
  release-service can still fail after the Release CR is accepted (signing, EC policy,
  registry push) — the gate polls `status.conditions[type=Released]` and reads the
  published digest from `status.artifacts.images`, marking the build `UNSTABLE` on
  failure/timeout.
- **Vars-file layout matches the existing `nightly.groovy` vs
  `3.19.groovy`/`4.21.groovy` convention** (`pipelines/vars/{foreman,katello}/`).
  `pipelines/vars/konflux/nightly.groovy` holds develop-stream data;
  `pipelines/release/pipelines/konflux-gate.groovy` and
  `pipelines/lib/konflux_gate.groovy` are stream-agnostic, so a future branched stream
  (e.g. 3.19) only needs a new vars file + job YAML.

## Prerequisite outside this repo

tenants-config needs a production `ReleasePlan` per Application (referenced via
`konflux_gate_release_plans`) that promotes stage → `quay.io/foreman/<image>:nightly`
on manual Release only, replacing today's single auto-release plan. RBAC is already
sufficient — `konflux-bot-0` is bound to `konflux-maintainer-user-actions`, which
already grants full CRUD on `snapshots`/`releases`/`releaseplans`.

## How this was validated locally

`./test theforeman.org` validates JJB rendering only — it never executes the Groovy —
so the actual gate mechanism was reproduced end-to-end by hand against the real latest
Konflux stage images, on a machine with Vagrant/libvirt and `oc` access to the
`theforeman-org-tenant` namespace:

1. Resolved the latest Snapshot per Application and its component image digests:
   ```
   oc get snapshot -n theforeman-org-tenant -l appstudio.openshift.io/application=candlepin \
     --sort-by=.metadata.creationTimestamp -o jsonpath='{.items[-1].metadata.name}'
   # -> candlepin-20260717-042540-000

   oc get snapshot candlepin-20260717-042540-000 -n theforeman-org-tenant \
     -o jsonpath='{range .spec.components[*]}{.name}{"\t"}{.containerImage}{"\n"}{end}'
   # -> candlepin-develop  quay.io/foreman/stage/candlepin@sha256:fd6c46d8ebfade535d636d01e296c5d5c17523d3e51dc0a784394f5218fcfebc
   ```
   (same for `foreman` — which bundles both `foreman-develop` and
   `foreman-proxy-develop` in one Snapshot — and `pulp`).

2. Confirmed the images are pullable without auth (`skopeo inspect docker://...`).

3. Cloned `theforeman/foremanctl` and ran its own setup:
   ```
   git clone https://github.com/theforeman/foremanctl.git
   cd foremanctl
   GITHUB_ACTIONS=true ./setup-environment
   ```

4. Patched `src/vars/images.yml` exactly as `konflux_gate_run_test` does — split each
   digest ref into an `image@sha256` / `<hex>` pair, since the deploy roles only join
   `"{{ x_container_image }}:{{ x_container_tag }}"` and there's no CLI/extra-vars
   override:
   ```diff
   -candlepin_container_image: quay.io/foreman/candlepin
   -candlepin_container_tag: "foreman-{{ container_tag_stream }}"
   -foreman_container_image: quay.io/foreman/foreman
   -foreman_container_tag: "{{ container_tag_stream }}"
   -foreman_proxy_container_image: "quay.io/foreman/foreman-proxy"
   -foreman_proxy_container_tag: "{{ container_tag_stream }}"
   +candlepin_container_image: quay.io/foreman/stage/candlepin@sha256
   +candlepin_container_tag: "fd6c46d8ebfade535d636d01e296c5d5c17523d3e51dc0a784394f5218fcfebc"
   +foreman_container_image: quay.io/foreman/stage/foreman@sha256
   +foreman_container_tag: "17345e73a80fcb251c7c16ae1bc9b57f595b6d48fd05167ae35854319d6d814a"
   +foreman_proxy_container_image: quay.io/foreman/stage/foreman-proxy@sha256
   +foreman_proxy_container_tag: "8453278c895cddae475dad3c7d413f9c8d6b1339590b3104373407e613a7a688"
    postgresql_container_image: quay.io/sclorg/postgresql-16-c10s
    postgresql_container_tag: "latest"
   -pulp_container_image: quay.io/foreman/pulp
   -pulp_container_tag: "foreman-{{ container_tag_stream }}"
   +pulp_container_image: quay.io/foreman/stage/pulp@sha256
   +pulp_container_tag: "9e83fa61b4c6c94d9432766ddffee5e24e38620d3cb24c75973e26792b7617a6"
    valkey_container_image: quay.io/sclorg/valkey-8-c10s
    valkey_container_tag: "latest"
   ```

5. Ran the VM lifecycle and deploy:
   ```
   ./forge vms start --vms "quadlet"
   ./forge setup-repositories
   ./foremanctl deploy --initial-admin-password=changeme --tuning development \
     --add-feature hammer --add-feature foreman-proxy --add-feature remote-execution
   ```
   Result: `PLAY RECAP ... ok=256 changed=160 failed=0`. Confirmed on the VM
   (`vagrant ssh quadlet -- sudo podman images`) that the real stage digests were
   pulled, not tags.

6. Ran the test suite:
   ```
   ./forge test --pytest-args="tests --ignore=tests/flavor --ignore=tests/feature/iop \
     --ignore=tests/backup_test.py --ignore=tests/certificate_bundle_test.py \
     --ignore=tests/feature/foreman/base_test.py \
     --deselect tests/feature/katello/client_test.py::test_foreman_content_view"
   ```
   Result: **270 passed**, including the full `candlepin_test.py` and `pulp_test.py`
   suites, `httpd`/`postgresql`/`valkey`/`foreman_target` health checks. The excluded
   items are documented inline in `konflux_gate_run_test` (pytest module-name collision
   in `tests/flavor`, tests needing `--add-feature iop` or a second Vagrant VM we don't
   provision here, and a few tests with a `./foremanctl`-relative-path bug unrelated to
   image content).

7. Verified `status.conditions`/`status.artifacts.images` shapes and the
   `release.appstudio.openshift.io/snapshot` label used for release idempotency
   directly against a real completed `Release` object on the cluster.

8. `vagrant destroy -f` to tear down.

## Test plan

- [x] `./test theforeman.org` passes (JJB render + `jenkins-lint.py` on all generated
      `config.xml`, including the new/changed jobs).
- [x] Local reproduction above validates the image-substitution mechanism and the full
      deploy+test flow against real Konflux stage images end-to-end.
- [ ] tenants-config MR (production ReleasePlan split) lands — external prerequisite,
      tracked separately.
- [ ] First real run on `ci.theforeman.org` once merged (draft PR pending that).

### DIFF
--- a/theforeman.org/pipelines/lib/konflux_gate.groovy
+++ b/theforeman.org/pipelines/lib/konflux_gate.groovy
@@ -1,0 +1,240 @@
+// Maps a Konflux Component name to the variable prefix foremanctl's
+// src/vars/images.yml uses for that component's image (confirmed against a real
+// checkout of theforeman/foremanctl): each component has a
+// `<prefix>_container_image` + `<prefix>_container_tag` pair which are joined as
+// "{{ x_container_image }}:{{ x_container_tag }}" by the deploy roles. There is no
+// CLI/extra-vars way to override these, so we sed-patch the checked-out file, same
+// mechanism theforeman/foreman-oci-images' own CI workflow uses to pin a build tag.
+def konflux_gate_image_var_prefix() {
+    return [
+        'candlepin-develop': 'candlepin',
+        'foreman-develop': 'foreman',
+        'foreman-proxy-develop': 'foreman_proxy',
+        'pulp-develop': 'pulp',
+    ]
+}
+
+// Last Snapshot for this Application that actually completed a Release
+// (`status.conditions[type=Released,status=True]`), used as the fallback image
+// source when a component's build times out this cycle. Relies on `jq` being
+// present on the 'el' agent.
+def konflux_gate_last_released_snapshot(app) {
+    return sh(
+        label: "resolve last released snapshot: ${app}",
+        script: """
+            ${konflux_oc_bin()} get release -n ${konflux_namespace()} -l appstudio.openshift.io/application=${app} -o json \
+              | jq -r '[.items[] | select(.status.conditions[]? | .type=="Released" and .status=="True")] | sort_by(.metadata.creationTimestamp) | last | .spec.snapshot // empty'
+        """,
+        returnStdout: true
+    ).trim()
+}
+
+// Polls for a new Snapshot for `app`. `previous` must be the snapshot captured by
+// katello.groovy *before* it triggered the rebuild (passed through as the
+// PREVIOUS_SNAPSHOTS build parameter) — not recomputed here, since by the time this
+// pipeline starts a fast build could already have landed, making "current latest"
+// indistinguishable from "the one we're supposed to wait for" and stalling this loop
+// until the timeout every single night. Returns [app, snapshot, stale] rather than
+// failing the build on timeout, so one slow/broken component doesn't block gating
+// the others (per-component timeout, not a whole-pipeline one).
+def konflux_gate_wait_for_new_snapshot(app, previous, timeoutMinutes) {
+    def resolved = previous
+    try {
+        timeout(time: timeoutMinutes, unit: 'MINUTES') {
+            waitUntil {
+                resolved = konflux_latest_snapshot(app)
+                return resolved && resolved != previous
+            }
+        }
+        return [app: app, snapshot: resolved, stale: false]
+    } catch (Exception ex) {
+        echo "Timed out waiting for a new Konflux Snapshot for '${app}' after ${timeoutMinutes}m; falling back to the last released Snapshot"
+        def fallback = konflux_gate_last_released_snapshot(app)
+        return [app: app, snapshot: fallback, stale: true]
+    }
+}
+
+def konflux_gate_image_ref(snapshot, component) {
+    return sh(
+        label: "resolve image ref: ${component}@${snapshot}",
+        script: "${konflux_oc_bin()} get snapshot ${snapshot} -n ${konflux_namespace()} -o jsonpath=\"{.spec.components[?(@.name=='${component}')].containerImage}\"",
+        returnStdout: true
+    ).trim()
+}
+
+// Release CRs carry a `release.appstudio.openshift.io/snapshot` label (confirmed on
+// the live cluster), so lookups don't need a jsonpath scan of every Release.
+def konflux_gate_find_existing_release(snapshot) {
+    return sh(
+        label: "check existing release: ${snapshot}",
+        script: "${konflux_oc_bin()} get release -n ${konflux_namespace()} -l release.appstudio.openshift.io/snapshot=${snapshot} -o jsonpath='{.items[0].metadata.name}'",
+        returnStdout: true
+    ).trim()
+}
+
+// Returns the created Release's actual name (generateName means we don't know it
+// up front). Uses `oc create`, not `apply` — generateName only makes sense as a
+// create-time directive (there's no existing name to diff against), and
+// find_existing_release() already guards against calling this when one exists.
+def konflux_gate_create_release(snapshot, releasePlan) {
+    // Matches the day/hour/minute-stamped naming Konflux itself uses for
+    // auto-created Snapshots/Releases (e.g. candlepin-20260713-224811-000), so a
+    // gate-created Release is identifiable at a glance and generateName's random
+    // suffix only has to disambiguate retries within the same minute.
+    def timestamp = sh(script: 'date -u +%Y%m%d-%H%M', label: 'timestamp release name', returnStdout: true).trim()
+    def release_yaml = """apiVersion: appstudio.redhat.com/v1alpha1
+kind: Release
+metadata:
+  generateName: ${snapshot}-gate-${timestamp}-
+  namespace: ${konflux_namespace()}
+spec:
+  snapshot: ${snapshot}
+  releasePlan: ${releasePlan}
+"""
+
+    writeFile(file: 'release.yaml', text: release_yaml)
+    try {
+        return sh(
+            label: "create release: ${snapshot}",
+            script: "${konflux_oc_bin()} create -f release.yaml -o jsonpath='{.metadata.name}'",
+            returnStdout: true
+        ).trim()
+    } finally {
+        sh(label: 'remove release manifest', script: 'rm -f release.yaml')
+    }
+}
+
+// Polls the Release CR's own `Released` condition rather than assuming success once
+// oc create returns — Konflux's release-service can still fail after acceptance
+// (signing, enterprise-contract policy, registry push errors), confirmed live on the
+// cluster: completed Releases carry
+// status.conditions[type=Released].{status,reason,message}. Returns null on timeout
+// (treated as "unknown", not "failed") so one slow release pipeline doesn't crash the
+// whole build.
+def konflux_gate_wait_for_release(releaseName, timeoutMinutes) {
+    def status = ''
+    try {
+        timeout(time: timeoutMinutes, unit: 'MINUTES') {
+            waitUntil {
+                status = sh(
+                    label: "poll release status: ${releaseName}",
+                    script: "${konflux_oc_bin()} get release ${releaseName} -n ${konflux_namespace()} -o jsonpath='{.status.conditions[?(@.type==\"Released\")].status}'",
+                    returnStdout: true
+                ).trim()
+                return status == 'True' || status == 'False'
+            }
+        }
+    } catch (Exception ex) {
+        echo "Timed out after ${timeoutMinutes}m waiting for Release '${releaseName}' to settle"
+        return null
+    }
+
+    def reason = sh(
+        label: "release reason: ${releaseName}",
+        script: "${konflux_oc_bin()} get release ${releaseName} -n ${konflux_namespace()} -o jsonpath='{.status.conditions[?(@.type==\"Released\")].reason}'",
+        returnStdout: true
+    ).trim()
+    def message = sh(
+        label: "release message: ${releaseName}",
+        script: "${konflux_oc_bin()} get release ${releaseName} -n ${konflux_namespace()} -o jsonpath='{.status.conditions[?(@.type==\"Released\")].message}'",
+        returnStdout: true
+    ).trim()
+
+    return [succeeded: status == 'True', reason: reason, message: message]
+}
+
+// The published, digest-pinned image(s) for a settled Release
+// (status.artifacts.images[].{name,shasum,urls}), confirmed live on the cluster. Only
+// meaningful once konflux_gate_wait_for_release() reports succeeded: true.
+def konflux_gate_release_artifacts(releaseName) {
+    def json = sh(
+        label: "release artifacts: ${releaseName}",
+        script: "${konflux_oc_bin()} get release ${releaseName} -n ${konflux_namespace()} -o jsonpath='{.status.artifacts.images}'",
+        returnStdout: true
+    ).trim()
+    return readJSON(text: json ?: '[]')
+}
+
+// Runs the real install/upgrade test against the staged images. Modeled directly on
+// centos.org/pipelines/foreman-discovery-image-build.groovy's Duffy usage: a single
+// root@duffy_box SSH session, no forklift/pipe-user layer. The Duffy pool used here
+// (metal-ec2-c5n-centos-9s-x86_64) is bare metal, so nested libvirt/KVM works the
+// same way it does for forklift's own Vagrant VMs today.
+//
+// imageRefs: Map of Konflux Component name -> digest-pinned containerImage ref
+// (e.g. "candlepin-develop": "quay.io/foreman/stage/candlepin@sha256:...").
+def konflux_gate_run_test(imageRefs) {
+    def boxname = 'duffy_box'
+    def var_prefix = konflux_gate_image_var_prefix()
+
+    setupDuffyClient()
+    provisionDuffy()
+
+    try {
+        stage('Prepare Duffy node') {
+            // Mirrors foreman-oci-images' integration.yml setup steps (libvirt/vagrant,
+            // Ansible, then foremanctl's own ./setup-environment) rather than inventing a
+            // new bootstrap sequence.
+            duffy_ssh('dnf install -y epel-release && dnf install -y libvirt vagrant qemu-kvm git python3 python3-pip make && systemctl enable --now libvirtd && vagrant plugin install vagrant-libvirt', boxname, './')
+            duffy_ssh('pip3 install --upgrade ansible-core', boxname, './')
+            duffy_ssh('git clone https://github.com/theforeman/foremanctl.git', boxname, './')
+            // GITHUB_ACTIONS=true keeps setup-environment on the "install into the
+            // system/host Python" path instead of creating a .venv it would activate for
+            // only that one SSH session — each duffy_ssh call is a fresh shell, so a venv
+            // activated here wouldn't be visible to the later forge/foremanctl calls.
+            duffy_ssh('cd foremanctl && GITHUB_ACTIONS=true ./setup-environment', boxname, './')
+
+            imageRefs.each { component, ref ->
+                def prefix = var_prefix[component]
+                if (!prefix) {
+                    error("konflux_gate_run_test: no foremanctl images.yml var prefix known for component '${component}'")
+                }
+
+                // images.yml only exposes "<image>:<tag>" (no digest-aware CLI/extra-vars
+                // override), so we split the digest ref and reassemble it across the two
+                // fields: image="<repo>@sha256", tag="<hex>" -> "<repo>@sha256:<hex>".
+                def parts = ref.tokenize('@')
+                def repo = parts[0]
+                def hex = parts[1].replace('sha256:', '')
+
+                duffy_ssh("sed -i 's#^${prefix}_container_image:.*#${prefix}_container_image: ${repo}@sha256#' foremanctl/src/vars/images.yml", boxname, './')
+                duffy_ssh("sed -i 's#^${prefix}_container_tag:.*#${prefix}_container_tag: \"${hex}\"#' foremanctl/src/vars/images.yml", boxname, './')
+            }
+        }
+
+        try {
+            stage('Deploy and test') {
+                duffy_ssh('cd foremanctl && ./forge vms start --vms "quadlet"', boxname, './')
+                duffy_ssh('cd foremanctl && ./forge setup-repositories', boxname, './')
+                duffy_ssh('cd foremanctl && ./foremanctl deploy --initial-admin-password=changeme --tuning development --add-feature hammer --add-feature foreman-proxy --add-feature remote-execution', boxname, './')
+                // Test scope confirmed by actually running `forge test` locally against
+                // the real latest stage digests (not guessed):
+                //  - tests/flavor/**: excluded, its own pulp_test.py collides with the
+                //    top-level tests/pulp_test.py under pytest's default rootdir-relative
+                //    import mode ("import file mismatch").
+                //  - tests/feature/iop/**: excluded, requires --add-feature iop which we
+                //    don't deploy here.
+                //  - tests/backup_test.py, tests/certificate_bundle_test.py,
+                //    tests/feature/foreman/base_test.py: excluded, they shell out to a
+                //    relative './foremanctl' and error with FileNotFoundError regardless
+                //    of image content — looks like a pytest-cwd issue in foremanctl
+                //    itself, not something specific to this gate.
+                //  - test_foreman_content_view: deselected, needs the separate "client"
+                //    Vagrant VM, which we don't start (only "quadlet").
+                // Everything else — including httpd/postgresql/valkey/foreman_target
+                // health checks and the full candlepin/pulp test files — ran and passed
+                // against the real Konflux stage images.
+                duffy_ssh('cd foremanctl && ./forge test --pytest-args="tests --ignore=tests/flavor --ignore=tests/feature/iop --ignore=tests/backup_test.py --ignore=tests/certificate_bundle_test.py --ignore=tests/feature/foreman/base_test.py --deselect tests/feature/katello/client_test.py::test_foreman_content_view"', boxname, './')
+            }
+        } catch (Exception ex) {
+            stage('Collect sos reports') {
+                duffy_ssh('cd foremanctl && ./forge sos', boxname, './')
+                duffy_scp('foremanctl/sos', "${env.WORKSPACE}/sos", boxname, './')
+                archiveArtifacts artifacts: 'sos/**', allowEmptyArchive: true
+            }
+            throw ex
+        }
+    } finally {
+        deprovisionDuffy()
+    }
+}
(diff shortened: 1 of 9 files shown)


theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,240 @@
...
+
+    setupDuffyClient()
+    provisionDuffy()
+
+    try {
+        stage('Prepare Duffy node') {
+            // Mirrors foreman-oci-images' integration.yml setup steps (libvirt/vagrant,
+            // Ansible, then foremanctl's own ./setup-environment) rather than inventing a
+            // new bootstrap sequence.
+            duffy_ssh('dnf install -y epel-release && dnf install -y libvirt vagrant qemu-kvm git python3 python3-pip make && systemctl enable --now libvirtd && vagrant plugin install vagrant-libvirt', boxname, './')
```
evgeni: Where do you get Vagrant and vagrant-libvirt from? That's not in EPEL.

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,240 @@
...
+            duffy_ssh('pip3 install --upgrade ansible-core', boxname, './')
```
evgeni: Surely we can just use ansible from CentOS stream here?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,240 @@
...
+                //    import mode ("import file mismatch").
+                //  - tests/feature/iop/**: excluded, requires --add-feature iop which we
+                //    don't deploy here.
+                //  - tests/backup_test.py, tests/certificate_bundle_test.py,
+                //    tests/feature/foreman/base_test.py: excluded, they shell out to a
+                //    relative './foremanctl' and error with FileNotFoundError regardless
+                //    of image content — looks like a pytest-cwd issue in foremanctl
+                //    itself, not something specific to this gate.
+                //  - test_foreman_content_view: deselected, needs the separate "client"
+                //    Vagrant VM, which we don't start (only "quadlet").
```
evgeni: Why not? You have a fat box, you can start the client

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,240 @@
...
+// (metal-ec2-c5n-centos-9s-x86_64) is bare metal, so nested libvirt/KVM works the
+// same way it does for forklift's own Vagrant VMs today.
+//
+// imageRefs: Map of Konflux Component name -> digest-pinned containerImage ref
+// (e.g. "candlepin-develop": "quay.io/foreman/stage/candlepin@sha256:...").
+def konflux_gate_run_test(imageRefs) {
+    def boxname = 'duffy_box'
+    def var_prefix = konflux_gate_image_var_prefix()
+
+    setupDuffyClient()
```
evgeni: This requires `CICO_API_KEY`, how do you obtain it?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,240 @@
...
+        script: "${konflux_oc_bin()} get release ${releaseName} -n ${konflux_namespace()} -o jsonpath='{.status.artifacts.images}'",
+        returnStdout: true
+    ).trim()
+    return readJSON(text: json ?: '[]')
+}
+
+// Runs the real install/upgrade test against the staged images. Modeled directly on
+// centos.org/pipelines/foreman-discovery-image-build.groovy's Duffy usage: a single
+// root@duffy_box SSH session, no forklift/pipe-user layer. The Duffy pool used here
+// (metal-ec2-c5n-centos-9s-x86_64) is bare metal, so nested libvirt/KVM works the
```
evgeni: this pipeline doesn't explicitly request this metal node, why is it mentioned here?

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Good catch — Vagrant/vagrant-libvirt aren't in EPEL, they're from COPR (forklift's own `roles/vagrant/tasks/main.yml` uses `pvalena/vagrant` and `evgeni/vagrant`). Fixed in fc02b08: moved node setup into a reusable playbook (`theforeman.org/ansible/setup_vagrant_libvirt.yml`) that adds those COPR repos and installs vagrant-libvirt as a prebuilt RPM instead of via `vagrant plugin install`.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fixed in fc02b08 — same playbook installs ansible-core via `dnf` from CentOS Stream's own repos instead of pip.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fair — fixed in fc02b08: `./forge vms start` with no `--vms` flag now, which defaults to `quadlet client` (forge's own default in `development/playbooks/vms/vms.yaml`). Verified locally that `test_foreman_content_view` passes with the client VM up.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Good question — it doesn't yet. Fixed in fc02b08 to bind it explicitly via `withCredentials([string(credentialsId: 'cico-api-key', ...)])` instead of relying on an ambient global env var (which only exists on ci.centos.org, not ci.theforeman.org where this job runs). The credential itself still needs to be created on ci.theforeman.org before this can run — that's a real prerequisite, tracking separately.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fair point — that comment overstated it. Fixed in fc02b08: reworded to make clear the pool choice comes from `provision_duffy.yml`'s own default, not something this pipeline requests or should claim credit/blame for.

theforeman.org/pipelines/release/pipelines/konflux-gate.groovy
```diff
@@ -0,0 +1,140 @@
...
+
+                    if (published) {
+                        currentBuild.description = ([currentBuild.description, 'Published:'] + published).minus(null).join('\n')
+                    }
+                }
+            }
+        }
+    }
+    post {
+        failure {
```
zjhuntin, line 136: Suggestion: release failures and timeouts set `currentBuild.result = 'UNSTABLE'` rather than `FAILURE`, so this notification won't fire for them. Adding an `unstable` block alongside might be worth considering:

```groovy
post {
    failure {
        notifyDiscourse(env, 'Konflux gate pipeline failed:', currentBuild.description)
    }
    unstable {
        notifyDiscourse(env, 'Konflux gate pipeline is unstable:', currentBuild.description)
    }
}
```

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,209 @@
...
+// *before* it triggers the rebuild (see PREVIOUS_SNAPSHOTS) — recomputing it here
+// would race with Konflux's own build latency. Per-component timeout, not
+// whole-pipeline, so one broken component doesn't block gating the others.
+def konflux_gate_wait_for_new_snapshot(app, previous, timeoutMinutes) {
+    def resolved = previous
+    try {
+        timeout(time: timeoutMinutes, unit: 'MINUTES') {
+            waitUntil {
+                resolved = konflux_latest_snapshot(app)
+                return resolved && resolved != previous
```
zjhuntin, line 36: Suggestion: if `oc get snapshot` throws a transient error (network blip, cluster hiccup), that exception propagates through `waitUntil` and `timeout`, lands in the outer `catch`, and silently triggers the stale-snapshot fallback as if a genuine timeout had occurred. Catching inside the closure would let it retry instead:

```groovy
waitUntil {
    try {
        resolved = konflux_latest_snapshot(app)
        return resolved && resolved != previous
    } catch (Exception ignored) {
        return false
    }
}
```

theforeman.org/pipelines/release/pipelines/katello.groovy
```diff
@@ -65,6 +65,40 @@ pipeline {
...
+            // konflux_gate_job_name job as a build parameter — not recomputed by that
+            // job once it starts. Konflux build+snapshot latency isn't bounded tightly
+            // enough to trust "whatever's latest when the gate job happens to start" as
+            // "the pre-rebuild snapshot": a fast build could already have landed by
+            // then, making previous == latest and stalling the gate's wait loop forever
+            // waiting for something newer that will never come.
+            when {
+                expression {
+                    try {
+                        konflux_components as boolean
```
zjhuntin, line 79: Suggestion: this guard checks `konflux_components`, but the stage body iterates `konflux_gate_applications`. Both are always defined together in the same vars file today, so it works fine in practice. Guarding on `konflux_gate_applications` instead might be slightly clearer about what the stage actually depends on, and safer if a future vars file ever defines one without the other.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fixed in b753431: transient errors from `konflux_latest_snapshot()` are now caught inside `waitUntil`, so the poll retries instead of treating them as a Snapshot timeout and falling back to a stale release.

theforeman.org/pipelines/release/pipelines/konflux-gate.groovy
Odilhao: Fixed in b753431: added `post { unstable { ... } }` to notify Discourse when a Release fails or times out.

theforeman.org/pipelines/release/pipelines/katello.groovy
Odilhao: Fixed in b753431: the `capture-konflux-snapshots` guard now checks `konflux_gate_applications`, which is the collection the stage iterates.

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,212 @@
+// Maps a Konflux Component name to the variable prefix in foremanctl's
+// src/vars/images.yml (<prefix>_container_image/_container_tag). No CLI/extra-vars
+// override exists, so we sed-patch the file instead.
+def konflux_gate_image_var_prefix() {
```
ehelms, line 1: Will this break for releases such as `foreman-5-0` ?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,212 @@
...
+        returnStdout: true
+    ).trim()
+
+    return [succeeded: status == 'True', reason: reason, message: message]
+}
+
+// The published, digest-pinned image(s) for a settled Release. Only meaningful once
+// konflux_gate_wait_for_release() reports succeeded: true.
+def konflux_gate_release_artifacts(releaseName) {
+    def json = sh(
```
ehelms, line 120: Do we know for sure this will return JSON and not go-style output? This came from an AI review wanting to double check:

```
 For an array of objects, oc/kubectl JSONPath normally renders Go-style output such as:                                                                                                       
                                                                                                                                                                                              
 ```text                                                                                                                                                                                      
   [map[name:... shasum:... urls:[...]]]                                                                                                                                                      
 ```                                                                                                                                                                                          
```

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,212 @@
...
+// box on ci.theforeman.org (no ci.centos.org delegation, no forklift).
+//
+// imageRefs: Map of Konflux Component name -> digest-pinned containerImage ref
+// (e.g. "candlepin-develop": "quay.io/foreman/stage/candlepin@sha256:...").
+def konflux_gate_run_test(imageRefs) {
+    def boxname = 'duffy_box'
+    def var_prefix = konflux_gate_image_var_prefix()
+
+    withCredentials([string(credentialsId: 'theforeman-duffy', variable: 'CICO_API_KEY')]) {
+        setupDuffyClient()
```
ehelms: This writes the credentials to a file in the home directory of the Jenkins node. I am not seeing anything that cleans that up afterward. I would think there is also a risk of two jobs trying to write to (or if clean up is added removing) the same file given it's being setup in the home directory.

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,212 @@
...
+    }
+    provisionDuffy()
```
ehelms: Should this be inside the try block to ensure the finally occurs?

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fixed in a7159f4. The 5.0 components now map to the same foremanctl image variables as their develop counterparts: candlepin, foreman, foreman_proxy, and pulp.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Confirmed against a completed foreman-5-0 Release in theforeman-org-tenant: this exact oc JSONPath expression emits valid JSON for status.artifacts.images, identical to the output of oc -o json piped to jq -c '.status.artifacts.images // []'. readJSON parses it successfully.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fixed in 15aec1d. Duffy setup, provisioning, the Ansible node setup, and deprovisioning now use a build-specific temporary `HOME`; it contains both the generated Duffy config and collection install, and is removed in a nested `finally`.

theforeman.org/pipelines/lib/konflux_gate.groovy
Odilhao: Fixed in 15aec1d. Provisioning now runs inside the outer `try`, so the `finally` attempts to deprovision a partially allocated session.

Odilhao:
@ehelms since we will have one full node on duffy when we run this type of pipeline, do you think we should expand the test matrix? if we add more tests like @evgeni mentioned to me in the past, how do we keep the test matrix update with the actual tests that we run on foremanctl?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+    return readJSON(text: json ?: '[]')
+}
+
+def konflux_gate_run_test(imageRefs) {
+    def boxname = 'duffy_box'
+    def var_prefix = konflux_gate_image_var_prefix()
+    def duffy_home = pwd(tmp: true)
+
+    try {
+        withEnv(["HOME=${duffy_home}"]) {
```
ehelms, line 134: Good adjustment -- we may need to extract this into a re-usable pattern later to ensure that the security is handled across any implementation.

ehelms:
> @ehelms since we will have one full node on duffy when we run this type of pipeline, do you think we should expand the test matrix? if we add more tests like @evgeni mentioned to me in the past, how do we keep the test matrix update with the actual tests that we run on foremanctl?

I wonder if to start with we should run an expansive set of tests (basically everything that we can in one go) and then expand to a matrix style of testing.

How big are the duffy nodes we are using?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+                def duffy_session = readFile(file: 'jenkins-jobs/centos.org/ansible/duffy_session')
+                runPlaybook(
+                    playbook: 'jenkins-jobs/theforeman.org/ansible/setup_vagrant_libvirt.yml',
+                    inventory: duffy_inventory('./'),
+                    limit: "duffy_session_${duffy_session}",
+                    options: ['-b'],
+                )
+            }
+
+            duffy_ssh('git clone https://github.com/theforeman/foremanctl.git', boxname, './')
```
ehelms, line 152: This will likely have to evolve to be able to test stable branches but can cross that bridge later.

Odilhao: Yes, that's part of the plan, we can make it more flexible after we get it working.

Odilhao:
> > @ehelms since we will have one full node on duffy when we run this type of pipeline, do you think we should expand the test matrix? if we add more tests like @evgeni mentioned to me in the past, how do we keep the test matrix update with the actual tests that we run on foremanctl?
> 
> I wonder if to start with we should run an expansive set of tests (basically everything that we can in one go) and then expand to a matrix style of testing.
> 
> How big are the duffy nodes we are using?

It's a small box with 192GB of RAM and 72vcpus

https://instances.vantage.sh/aws/ec2/c5n.metal?currency=USD

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+        label: "resolve image ref: ${component}@${snapshot}",
+        script: "${konflux_oc_bin()} get snapshot ${snapshot} -n ${konflux_namespace()} -o jsonpath=\"{.spec.components[?(@.name=='${component}')].containerImage}\"",
+        returnStdout: true
+    ).trim()
+}
+
+def konflux_gate_find_existing_release(snapshot) {
+    return sh(
+        label: "check existing release: ${snapshot}",
+        script: "${konflux_oc_bin()} get release -n ${konflux_namespace()} -l release.appstudio.openshift.io/snapshot=${snapshot} -o jsonpath='{.items[0].metadata.name}'",
```
zjhuntin: On the first promotion of a Snapshot, this label query returns an empty `.items`; JSONPath index `[0]` can fail on the empty list, so the `sh` step aborts before the `else` branch can create a Release. Could this lookup treat no match as an empty string (for example, parse `oc -o json` with `jq -r '.items[0].metadata.name // empty'`) so first-time releases proceed to `konflux_gate_create_release`?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+                duffy_ssh("sed -i 's#^${prefix}_container_tag:.*#${prefix}_container_tag: \"${hex}\"#' foremanctl/src/vars/images.yml", boxname, './')
+            }
+        }
+
+        try {
+            stage('Deploy and test') {
+                duffy_ssh('cd foremanctl && ./forge vms start', boxname, './')
+                duffy_ssh('cd foremanctl && ./forge setup-repositories', boxname, './')
+                duffy_ssh('cd foremanctl && ./foremanctl deploy --initial-admin-password=changeme --tuning development --add-feature hammer --add-feature foreman-proxy --add-feature remote-execution', boxname, './')
+                duffy_ssh('cd foremanctl && ./forge test --pytest-args="--ignore=tests/flavor --ignore=tests/feature/iop -k \'not test_foreman_initial_organization and not test_foreman_initial_location\'"', boxname, './')
```
zjhuntin: `duffy_ssh` interpolates this command inside a local single-quoted shell argument. The `\'` escapes become literal single quotes in the Groovy string, so the local shell closes its quote here and splits the expression before SSH; the remote `forge test` no longer receives the `-k` selector as one expression and can fail or run with unintended selection. Could the SSH wrapper safely quote remote commands or otherwise preserve this selector as a single argument?

theforeman.org/pipelines/lib/konflux_gate.groovy
```diff
@@ -0,0 +1,194 @@
...
+}
+
+def konflux_gate_wait_for_new_snapshot(app, previous, timeoutMinutes) {
+    def resolved = previous
+    try {
+        timeout(time: timeoutMinutes, unit: 'MINUTES') {
+            waitUntil {
+                try {
+                    resolved = konflux_latest_snapshot(app)
+                    return resolved && resolved != previous
```
zjhuntin: Any changed Application Snapshot is accepted, but the `foreman` Application maps to both `foreman-develop` and `foreman-proxy-develop`, which are triggered separately. Konflux can create an Application Snapshot after one component build succeeds, using that new image and the sibling's last-promoted image ([Snapshot preparation](https://github.com/konflux-ci/integration-service/blob/main/gitops/snapshot.go#L1015-L1071)). If that Snapshot becomes latest before the sibling build completes, this gate tests/releases a mixed set and misses the sibling's new Snapshot. Could this wait until a Snapshot contains refreshed images for every mapped component, rather than accepting any name change?

zjhuntin (changes requested):
Three actionable issues block the gate from reliably testing and promoting the intended image set.

ehelms:
> It's a small box with 192GB of RAM and 72vcpus
> 
> https://instances.vantage.sh/aws/ec2/c5n.metal?currency=USD

Is that all? :grin:

PR #601, opened 2026-08-11
Foreman 5.0 branching

ogajduse:


zjhuntin (approved):
Looks good, no el10 in those installs/upgrades! 😄

